PR #2, opened 2018-05-21
fetch poi in elasticsearch

antoine-de:
follow #1 

fetch the poi in mimir's elasticsearch.

add some test backed by docker.

### DIFF
--- a/api/pois.py
+++ b/api/pois.py
@@ -1,0 +1,16 @@
+
+
+def fetch_es_poi(id, lang) -> dict:
+    # TODO!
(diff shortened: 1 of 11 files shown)


api/pois.py
```diff
@@ -0,0 +1,26 @@
...
+    poi = {
+        'name': es_poi['name'],
+        'id': es_poi['id'],
+        'label': es_poi['label'],
+    }
+    # TODO!
+    return poi
+
+
+def get_poi(id, lang):
```
amatissart: What if `lang` parameter is absent in the HTTP query ? I would prefer an explicit default.

PR #3, opened 2018-05-23
add 3 ways to configure the API

antoine-de:
The configuration can be given from different ways:
1. a default settings is available in utils/default_settings.yaml
2. a yaml settings file can be given with an env var IDUNN_CONFIG_FILE
    (the default settings is still loaded and overriden)
3. specific variable can be overriden with env var. They need to be
given like "IDUNN_{var_name}={value}"
    eg IDUNN_ES_URL=...


app.py
```diff
@@ -1,12 +1,17 @@
 from apistar import App, Include
 from api.urls import api_urls
+from utils.settings import SettingsComponent
```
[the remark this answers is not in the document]
antoine-de: oups :roll_eyes:

utils/settings.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def _load_specific_config(self, project_name):
+        """
+        load settings from a yaml file whose pth in given
+        with the env var {project}_CONFIG_FILE (eg. IDUNN_CONFIG_FILE for the IDUNN project)
+        """
+        path = os.environ.get(f'{project_name}_CONFIG_FILE')
+        if not path:
+            return
+        config_path = path
```
amatissart: unnecessary variable

utils/settings.py
```diff
@@ -0,0 +1,59 @@
...
+        self.settings.update(_load_yaml_file(config_path))
+
+    def _load_from_env_var(self, project_name):
+        """
+        overrides existing settings from the env var.
+        The env var need to be defined as {project}_{var_name}
+        eg. IDUNN_ES to override the "ES" settings in the IDUNN project.
+        """
+        for k in self.settings.keys():
+            ovverided_value = os.environ.get(f'{project_name}_{k}')
```
amatissart: typo: ovverided :arrow_right: overridden

PR #6, opened 2018-05-30
Add fields - QMAPS-303

jbgriesner:
This PR adds a basic object model structure for blocks returned by the API.
Simple tests have been added (note the json in the fixture corresponds to the bragi answer, but does not corresponds to the es answer that idunn makes).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
   - pipenv install --dev
 
 script:
-  - pipenv run pytest tests
+  - pipenv run pytest
 
 notifications:
   email: false
--- a/idunn/api/poi.py
+++ b/idunn/api/poi.py
@@ -1,0 +1,94 @@
+from apistar import types, validators
+from idunn.blocks.phone import PhoneBlock
+from idunn.blocks.opening_hour import OpeningHourBlock
+
+BLOCKS_ORDER = [
+    OpeningHourBlock,
+    PhoneBlock
+]
+
+def build_blocks(es_poi, lang):
+    return [c.from_es(es_poi, lang) for c in BLOCKS_ORDER]
+
+def get_name(properties, lang):
+    """Return the POI name from the properties field of the elastic response
+    Here 'name' corresponds to the POI name in the language of the user request (i.e. 'name:{lang}' field).
+
+    If lang is None or if name:lang is not in the properties
+    Then name receives the local name value
+
+    'local_name' corresponds to the name in the language of the country where the POI is located.
+
+    >>> get_name({}, 'fr') is None
+    True
+
+    >>> get_name({'name':'spontini', 'name:en':'spontinien', 'name:fr':'spontinifr'}, None)
+    'spontini'
+
+    >>> get_name({'name':'spontini', 'name:en':'spontinien', 'name:fr':'spontinifr'}, 'cz')
+    'spontini'
+
+    >>> get_name({'name':'spontini', 'name:en':'spontinien', 'name:fr':'spontinifr'}, 'fr')
+    'spontinifr'
+    """
+    name = properties.get(f'name:{lang}')
+    if name is None:
+        name = properties.get('name')
+    return name
+
+def get_geom(es_poi):
+    """Return the correct geometry from the elastic response
+
+    A correct geometry means both lat and lon coordinates are required
+
+    >>> get_geom({}) is None
+    True
+
+    >>> get_geom({'coord':{"lon": None, "lat": 48.858260156496016}}) is None
+    True
+
+    >>> get_geom({'coord':{"lon": 2.2944990157640612, "lat": None}}) is None
+    True
+
+    >>> get_geom({'coord':{"lon": 2.2944990157640612, "lat": 48.858260156496016}})
+    {'coordinates': [2.2944990157640612, 48.858260156496016], 'type': 'Point'}
+    """
+    geom = None
+    if 'coord' in es_poi:
+        coord = es_poi.get('coord')
+        lon = coord.get('lon')
+        lat = coord.get('lat')
+        if lon is not None and lat is not None:
+            geom = {
+                'coordinates': [lon, lat],
+                'type': 'Point'
+            }
+    return geom
+
+class POI(types.Type):
+    id = validators.String(allow_null=True)
+    name = validators.String(allow_null=True)
+    local_name = validators.String(allow_null=True)
+    class_name = validators.String(allow_null=True)
+    subclass_name = validators.String(allow_null=True)
+    geometry = validators.Object(allow_null=True)
+    address = validators.Object(allow_null=True)
+    blocks = validators.Array(items=BLOCKS_ORDER)
+
+    @classmethod
+    def load_poi(cls, es_poi, lang):
+
+        properties = es_poi.get('properties', {})
+
+        return cls(
+            id=es_poi['id'],
+            name=get_name(properties, lang),
+            local_name=properties.get('name'),
+            class_name=properties.get('poi_class'),
+            subclass_name=properties.get('poi_subclass'),
+            geometry=get_geom(es_poi),
+            address= {
+                'label': es_poi.get('address').get('label')
+            },
+            blocks=build_blocks(es_poi, lang)
+        )
--- a/idunn/blocks/opening_hour.py
+++ b/idunn/blocks/opening_hour.py
@@ -1,0 +1,28 @@
+import re
+import datetime
+from apistar import types, validators
+
+class OpeningHourBlock(types.Type):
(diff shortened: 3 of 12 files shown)


idunn/blocks/opening_hour.py
```diff
@@ -86,14 +66,17 @@ def from_es(cls, es_poi, lang):
...
 
         ohr = hoh.render_field(clean_raw, universal=False)
-        description = ohr.full_description()
-        days, is247 = from_desc(description)
 
-        status = ''
+        # plaintext_week = ohr.plaintext_week_description()
+
+        # try:
+        #    days = [{'day_name' d.split(':', 1)[0]: d.split(':', 1)[1].strip()} for d in plaintext_week.splitlines()]
```
antoine-de: exception should be use only sparcingly. check `next`

(and I fhink we should clean this if not used

idunn/blocks/opening_hour.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @classmethod
+    def from_es(cls, es_poi, lang):
+        logger = logging.getLogger('idunn')
+        is247 = False
+
+        raw = es_poi.get('properties', {}).get('opening_hours')
+        if raw is None:
+            return None
+        elif raw is '24/7':
```
antoine-de: you can rewrite this: 
```python
if raw is None:
    return None
is247  = raw == '24/7'
```
(for this use it seems more appropriate to use `==` than `is`)

idunn/blocks/opening_hour.py
```diff
@@ -0,0 +1,77 @@
...
+            is247 = True
+
+        try:
+            clean_raw = hoh.OHParser.sanitize(raw)
+            oh = hoh.OHParser(clean_raw)
+        except (SpanOverMidnight, ParseError) as err:
+            logger.error(f"OpeningHour error: {err}")
+            return None
+
+        status = 'close'
```
antoine-de: minor but you can write this `status = 'open' if oh.is_open() else 'close'`

 do we want to use the `open()` function of the library  ? what time is used to compute this ? the server's ? the poi's ?

idunn/blocks/opening_hour.py
```diff
@@ -0,0 +1,77 @@
...
+            type='OpeningHourBlock',
+            status=status,
+            next_transition_time=ntt,
+            seconds_before_next_transition=sbnt,
+            is_24_7=is247,
+            raw=clean_raw,
+            days=[] # format definition left for future work
+        )
+
+if __name__ == "__main__":
```
antoine-de: hum why do you need this ?
we should use pytest to run the doc test, not mains

tests/test_api.py
```diff
@@ -33,12 +33,18 @@ def test_basic_query(orsay_museum):
...
+
+    assert resp['id'] == 'osm:way:63178753'
+    assert resp['name'] == "Musee d'Orsay"
+    assert resp['local_name'] == "Musée d'Orsay"
+    assert resp['class_name'] == 'museum'
+    assert resp['subclass_name'] == 'museum'
+    assert resp['address']['label'] == '62B Rue de Lille (Paris)'
+    assert resp['blocks'][0]['btype'] == 'OpeningHourBlock'
+    assert resp['blocks'][0]['is_24_7'] == False
+    assert resp['blocks'][0]['days'][2] == 'Wednesday'
```
antoine-de: since we use the `lang`, it would be nice to have a test with a `lang` param

idunn/blocks/opening_hour.py
jbgriesner: done

idunn/blocks/opening_hour.py
jbgriesner: done

tests/test_api.py
jbgriesner: done

idunn/blocks/opening_hour.py
jbgriesner: They use the present time of the server:
dt = datetime.datetime.now().replace(tzinfo=pytz.UTC)
and check if the corresponding moment belongs to periods of the POI:
https://github.com/rezemika/humanized_opening_hours/blob/63a484ad76a57276823f04240c4ed1af6a516ec6/humanized_opening_hours/temporal_objects.py#L55

I thought it was ok to let investigations of better alternatives as a future work.

idunn/api/poi.py
```diff
@@ -0,0 +1,53 @@
...
+                geom = {
+                    'coordinates': [lon, lat],
+                    'type': 'Point'
+                }
+
+        local_name = es_poi.get('properties', {}).get('name')
+        if lang is None:
+            name = local_name
+        else:
+            name = es_poi.get('properties', {}).get(f'name:{lang}')
```
antoine-de: it does not work if the lang is not in the name.

maybe you should wrap this in a function, this way you can add some doc test on the corner cases (and some comments)

tests/test_api.py
```diff
@@ -33,11 +33,29 @@ def test_basic_query(orsay_museum):
...
+    client = TestClient(app)
+    response = client.get(
+        url=f'http://localhost/v1/pois/{orsay_museum}?lang=it',
+    )
+
+    assert response.status_code == 200
+
+    resp = response.json()
+
+    assert resp['name'] == "Museo d'Orsay"
```
antoine-de: I think it's nice to test all the field there too

tests/test_api.py
```diff
@@ -54,6 +72,7 @@ def test_unknow_poi(orsay_museum):
 
 def test_schema():
```
antoine-de: you can comment all the test

idunn/blocks/opening_hour.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def from_es(cls, es_poi, lang):
+        logger = logging.getLogger('idunn')
+
+        raw = es_poi.get('properties', {}).get('opening_hours')
+        is247 = False
+        if raw is None:
+            return None
+        elif raw == '24/7':
```
antoine-de: you haven't changed this from my earlier comment

idunn/blocks/opening_hour.py
```diff
@@ -0,0 +1,34 @@
...
+            is247 = True
+
+        return cls(
+            type='OpeningHourBlock',
```
antoine-de: can't we get this from `cls` ?

idunn/blocks/block.py
```diff
@@ -0,0 +1,9 @@
+from apistar import types, validators
+
+class Block(types.Type):
```
antoine-de: this class is worthless now no ?

idunn/blocks/opening_hour.py
amatissart: Also, the class name should not be used as `type`.
`opening_hours` is expected here in snake case (see the specs).

.travis.yml
```diff
@@ -12,6 +12,7 @@ install:
 
 script:
   - pipenv run pytest tests
+  - pipenv run pytest --doctest-modules
```
antoine-de: I think you can acheive this by adding a configuration file in the project's root. 

this way the doctest will always be run, even in local

idunn/api/poi.py
```diff
@@ -0,0 +1,91 @@
...
+
+    'local_name' corresponds to the name in the language of the country where the POI is located.
+
+    >>> get_name({}, 'fr') is None
+    True
+
+    >>> get_name({'name':'spontini', 'name:en':'spontinien', 'name:fr':'spontinifr'}, None)
+    'spontini'
+
+    >>> get_name({'name':'spontini', 'name:en':'spontinien', 'name:fr':'spontinifr'}, 'fr')
```
antoine-de: a test with a name a an unknown lang would be nice

idunn/blocks/opening_hour.py
```diff
@@ -0,0 +1,31 @@
...
+    status = validators.String()
+    next_transition_time = validators.String()
+    seconds_before_next_transition = validators.Integer()
+    is_24_7 = validators.Boolean()
+    raw = validators.String()
+    days = validators.Array()
+
+    @classmethod
+    def from_es(cls, es_poi, lang):
+        logger = logging.getLogger('idunn')
```
antoine-de: the logger is not used

.travis.yml
```diff
@@ -11,7 +11,7 @@ install:
   - pipenv install --dev
 
 script:
-  - pipenv run pytest tests
+  - pipenv run pytest
```
antoine-de: you need to update travis

antoine-de: my bad, I missread

PR #8, opened 2018-06-06
don't add null blocks to output

nlehuby:
This fix the `Server error` when some block is null.
Not sure this is the correct way to handle it though :thinking:

### DIFF
--- a/idunn/api/poi.py
+++ b/idunn/api/poi.py
@@ -8,7 +8,12 @@ BLOCKS_ORDER = [
 ]
 
 def build_blocks(es_poi, lang):
-    return [c.from_es(es_poi, lang) for c in BLOCKS_ORDER if c.from_es(es_poi, lang)]
+    blocks = []
+    for c in BLOCKS_ORDER:
+        block = c.from_es(es_poi, lang)
+        if block is not None:
+            blocks.append(block)
+    return blocks
 
 def get_name(properties, lang):
     """Return the POI name from the properties field of the elastic response
--- a/idunn/api/poi.py
+++ b/idunn/api/poi.py
@@ -8,7 +8,7 @@ BLOCKS_ORDER = [
 ]
 
 def build_blocks(es_poi, lang):
-    return [c.from_es(es_poi, lang) for c in BLOCKS_ORDER]
+    return [c.from_es(es_poi, lang) for c in BLOCKS_ORDER if c.from_es(es_poi, lang)]
 
 def get_name(properties, lang):
     """Return the POI name from the properties field of the elastic response
@@ -73,7 +73,7 @@ class POI(types.Type):
     subclass_name = validators.String(allow_null=True)
     geometry = validators.Object(allow_null=True)
     address = validators.Object(allow_null=True)
-    blocks = validators.Array(items=BLOCKS_ORDER)
+    blocks = validators.Array()
 
     @classmethod
     def load_poi(cls, es_poi, lang):
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -4,6 +4,20 @@ import pytest
 import json
 import os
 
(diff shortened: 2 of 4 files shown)


idunn/api/poi.py
```diff
@@ -8,7 +8,7 @@
 ]
 
 def build_blocks(es_poi, lang):
-    return [c.from_es(es_poi, lang) for c in BLOCKS_ORDER]
+    return [c.from_es(es_poi, lang) for c in BLOCKS_ORDER if c.from_es(es_poi, lang)]
```
antoine-de: this will create twice the poi, I think it's better to filter the none afterward (or use a normal loop :stuck_out_tongue_winking_eye: )

tests/test_api.py
```diff
@@ -26,6 +26,19 @@ def orsay_museum(es_client):
         return poi_id
 
 
+@pytest.fixture(scope="module")
+def blancs_manteaux(es_client):
```
antoine-de: I think this can be factorized with the method above

tests/test_api.py
```diff
@@ -66,6 +79,25 @@ def test_lang(orsay_museum):
     assert resp['blocks'][1]['type'] == 'phone'
     assert resp['blocks'][0]['is_24_7'] == False
 
+def test_block_null(blancs_manteaux):
+    client = TestClient(app)
+    response = client.get(
+        url=f'http://localhost/v1/pois/{blancs_manteaux}?lang=it',
```
antoine-de: by curiosity we test with `lang=it` for a reason ?

tests/test_api.py
```diff
@@ -66,6 +79,25 @@ def test_lang(orsay_museum):
     assert resp['blocks'][1]['type'] == 'phone'
     assert resp['blocks'][0]['is_24_7'] == False
 
+def test_block_null(blancs_manteaux):
```
antoine-de: It's nice to add a block comment explaining what you will test

idunn/api/poi.py
```diff
@@ -73,7 +78,7 @@ class POI(types.Type):
     subclass_name = validators.String(allow_null=True)
     geometry = validators.Object(allow_null=True)
     address = validators.Object(allow_null=True)
-    blocks = validators.Array(items=BLOCKS_ORDER)
+    blocks = validators.Array()
```
antoine-de: why do you removed this ?
It's not a big problem as the schema is currently broken, but in the futur, we'll want to reactive this

tests/test_api.py
```diff
@@ -14,17 +28,14 @@ def orsay_museum(es_client):
...
-                        refresh=True)
-        return poi_id
+    return load_poi('orsay_museum.json', es_client)
 
+@pytest.fixture(scope="module")
+def blancs_manteaux(es_client):
+    """
+    fill elasticsearch with one poi, the church des blancs manteaux
+    """
+    return load_poi('blancs_manteaux.json', es_client)
```
antoine-de: this breaks test encapsulation as now If you run only the test `test_block_null` the test won't pass.

To fix this, you need to create:

```python 
@pytest.fixture(scope="module")
def init_indicies(es_client):
   .. stuff from above
```

and inject it in `orsay_museum` and `blanc_manteau`

jbgriesner: Wops you'r right, thanks!

tests/test_api.py
```diff
@@ -4,6 +4,12 @@
 import json
 import os
 
+@pytest.fixture(scope="module")
```
antoine-de: this works without the autouse ?
I though you needed either an autouse=True or to inject the fixture in another one. :thinking:

jbgriesner: Yes it works for each test individually and also all tests together.

PR #9, opened 2018-06-06
Phone - handle multiple tags

nlehuby:
Use `phone` or `contact:phone` tag to fill the `phone` block


idunn/blocks/phone.py
```diff
@@ -10,7 +10,10 @@ class PhoneBlock(types.Type):
     def from_es(cls, es_poi, lang):
         raw = es_poi.get('properties', {}).get('phone')
         if raw is None:
-            return None
+            raw = es_poi.get('properties', {}).get('contact:phone')
```
antoine-de: completly minor, but you can also write this:
```python
raw = es_poi.get('properties', {}).get('phone') or es_poi.get('properties', {}).get('contact:phone')
if raw is None:
    return None
```
(maybe with a temporary variable for `properties`

jbgriesner:
:+1:

jbgriesner:
Done.

PR #11, opened 2018-06-12
Implement "wikipedia" block

amatissart:
This defines a new `wikipedia` block, nested in an `information` block.

* Wikipedia summary is fetched using [Wikimedia REST API](https://en.wikipedia.org/api/rest_v1/#!/Page_content/get_page_summary_title)
* We look for an article title in the `wikipedia` tag. If the tag value does not match the requested language, the corresponding title is requested beforehand (using https://www.mediawiki.org/wiki/API:Langlinks)
* `wikidata` tag is never used at this point 

---

Note that a new `BlocksValidator` is defined to validate each block in responses, based on its type.


idunn/utils/default_settings.yaml
```diff
@@ -1,3 +1,6 @@
 
 ES_URL: http://localhost:9200/
 
+WIKI_USER_AGENT: 'Idunn' # Used in requests to external wiki* APIs
+
+DEFAULT_LANGUAGE: 'en' # Fallback when no 'lang' in request
```
jbgriesner: So are we sure we don't want to force to get the lang ?

idunn/blocks/information.py
```diff
@@ -0,0 +1,123 @@
...
+        except RequestException:
+            logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
+    return wrapper
+
+
+class WikipediaSession:
+    _session = None
+    timeout = 1. # seconds
+
+    api_v1_base_pattern = "https://{lang}.wikipedia.org/api/rest_v1"
```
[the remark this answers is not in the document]
amatissart: Sure, why not. I suppose we can treat them as constants.
Even if the difference between settings, constants, and default values is always a bit fuzzy for me ^^

idunn/utils/default_settings.yaml
amatissart: I am not convinced myself. This defines a default value upstream in the handler function, so that blocks implementations no longer need to handle None in `lang`.  
But we might need other language fallbacks in the future, when we'll derive the language from HTTP headers (that may contain multiple languages)

idunn/blocks/base.py
```diff
@@ -0,0 +1,56 @@
+from apistar import types, validators
+
+class BaseBlock(types.Type):
+    BLOCK_TYPE = '' # To override in each subclass
+
+    type = validators.String()
+
+    def __init__(self, *args, **kwargs):
+        if not args:
+            if not self.BLOCK_TYPE:
```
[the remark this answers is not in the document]
antoine-de: I don't think it's worth it, that's fine for me

tests/test_api_with_wiki.py
```diff
@@ -0,0 +1,88 @@
...
+        'type': 'information',
+        'blocks': [
+            {
+                'type': 'wikipedia',
+                'url': 'https://es.wikipedia.org/wiki/Museo_del_Louvre',
+                'title': 'Museo del Louvre',
+                'description': 'El Museo del Louvre es el museo nacional de Francia ...'
+            }
+        ]
+    }
```
antoine-de: missing newline

PR #12, opened 2018-06-14
Opening hours block 

jbgriesner:
This PR defines the opening hours block:
- the osm opening hours are parsed with the [humanized_opening_hours](https://github.com/rezemika/humanized_opening_hours)  library.
- timezones are managed with `pytz` and `tzwhere` libraries.
- tests use `freezegun` library to mock the datetime.

TODO: the format for the `days` field has to be defined.


jbgriesner:
With [the new hoh release](https://github.com/rezemika/humanized_opening_hours/issues/11) idunn can now parse the louvre opening hours.

tests/test_opening_hours.py
```diff
@@ -0,0 +1,38 @@
...
+    assert response == OpeningHourBlock(
+        status='open',
+        next_transition_time='2018-06-14T22:00:00+00:00',
+        seconds_before_next_transition=45000,
+        is_24_7=False,
+        raw='Mo-Su 10:00-22:00',
+        days=[]
+    )
+
+    freezer.stop()
```
antoine-de: I don't think you should do it like this as the tests will now have side effects while failing.

rather use context manager or decorator

tests/test_opening_hours.py
```diff
@@ -0,0 +1,38 @@
...
+                "opening_hours": "Mo-Su 10:00-22:00"
+            }
+        },
+        lang='en'
+    )
+
+    assert response == OpeningHourBlock(
+        status='open',
+        next_transition_time='2018-06-14T22:00:00+00:00',
+        seconds_before_next_transition=45000,
```
antoine-de: the test is difficult to read as it is, please add meaningfull comments.

and I don't understand, if it's `9:30` and the shop open as `10:00` how is it open ?

and open question, the `next_transition_time` seems to be formated as UTC (but it's local no ?). should we display it in local, in UTC or in the timezone of the POI ? (I'd say the timezone of the poi)

amatissart: it's supposed to be 9:30 UTC (11:30 in Moscow), so the shop is open.
However the next_transition should be localized in Moscow time (22:00 local time, so the prefix "+00:00" is wrong here). Also the field should be named `next_transition_datetime`.

Another problem that is probably related, the seconds_before_next_transition value is wrong. If it's 11:30 local time, the shop closes at 22:00 (in 10.5 hours = 37800s)

jbgriesner: done.
I think `next_transition_time` should be displayed in UTC.

tests/test_opening_hours.py
jbgriesner: done

tests/test_opening_hours.py
```diff
@@ -0,0 +1,47 @@
...
+            {
+                "coord": {
+                    "lon": 37.588161523500276,
+                    "lat": 55.74831406552745
+                },
+                "properties": {
+                    "opening_hours": "Mo-Su 10:00-22:00"
+                }
+            },
+            lang='en'
```
antoine-de: do we need the lang ?

idunn/blocks/opening_hour.py
```diff
@@ -20,11 +57,36 @@ def from_es(cls, es_poi, lang):
...
+            logging.info("OSM opening_hour field cannot span over midnight", exc_info=True)
+            return None
+
+        status = 'close'
+        next_transition_datetime = ''
+        time_before_next = 0
+        poi_tz = get_tz(es_poi)
+        if poi_tz is not None:
+            poi_dt = UTC.localize(datetime.utcnow()).astimezone(poi_tz)
+            if poi_dt is not None:
```
antoine-de: how would this be none ?

tests/test_opening_hours.py
jbgriesner: Yes since 5971d611d6c1a1df7f9eaabd9ab6eb8110496f45 it's mandatory.

idunn/blocks/opening_hour.py
```diff
@@ -1,13 +1,50 @@
...
+        lon = coord.get('lon')
+        lat = coord.get('lat')
+        if lon is not None and lat is not None:
+            return timezone(tz.tzNameAt(lat, lon))
+    return None
 
 class OpeningHourBlock(BaseBlock):
     BLOCK_TYPE = 'opening_hours'
 
     status = validators.String()
```
amatissart: I would enumerate all possible status here (using `enum` param: https://docs.apistar.com/api-guide/type-system/#string)

And contrary to the original snippet, I think "open" and "closed" (instead of "close") would be more correct in proper english.

jbgriesner: done

PR #15, opened 2018-07-11
Circuit breaker

jbgriesner:
Prevent too many calls to external services when they are offline.
After 15 fails we stop to invoke wikipedia api for 120 seconds.

### DIFF
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -5,19 +5,42 @@ from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
 import pybreaker
 
-wiki_breaker = pybreaker.CircuitBreaker(fail_max=15, reset_timeout=120)
+class HTTPError40X(HTTPError):
+    pass
 
-def handle_requests_error(f):
-    def wrapper(*args, **kwargs):
-        try:
-            return wiki_breaker.call(f,*args, **kwargs)
-        except HTTPError:
-            logging.info("Got HTTP error in {}".format(f.__name__), exc_info=True)
-        except Timeout:
-            logging.warning("External API timed out in {}".format(f.__name__), exc_info=True)
-        except RequestException:
-            logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
-    return wrapper
+class WikipediaBreaker:
+    _breaker = None
+
+    @classmethod
+    def init_breaker(cls):
+        from app import settings
+        cls._breaker = pybreaker.CircuitBreaker(
+                fail_max=settings['CIRCUIT_FAILMAX'],
+                reset_timeout=settings['CIRCUIT_TIMEOUT'],
+                exclude=[HTTPError40X])
+
+    @classmethod
+    def get_breaker(cls):
+        if cls._breaker is None:
+            cls.init_breaker()
+        return cls._breaker
+
+    @classmethod
+    def handle_requests_error(cls, f):
+        def wrapper(*args, **kwargs):
+            if cls._breaker is None:
+                cls.init_breaker()
+            try:
+                return cls._breaker(f)(*args, **kwargs)
+            except pybreaker.CircuitBreakerError:
+                logging.info("Got CircuitBreakerError in {}".format(f.__name__), exc_info=True)
+            except HTTPError:
+                logging.info("Got HTTP error in {}".format(f.__name__), exc_info=True)
+            except Timeout:
+                logging.warning("External API timed out in {}".format(f.__name__), exc_info=True)
+            except RequestException:
+                logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
+        return wrapper
 
 
 class WikipediaSession:
@@ -36,16 +59,18 @@ class WikipediaSession:
             self._session.headers.update({"User-Agent": user_agent})
         return self._session
 
-    @handle_requests_error
+    @WikipediaBreaker.handle_requests_error
     def get_summary(self, title, lang):
         url = "{base_url}/page/summary/{title}".format(
             base_url=self.API_V1_BASE_PATTERN.format(lang=lang), title=title
         )
         resp = self.session.get(url=url, params={"redirect": True}, timeout=self.timeout)
+        if resp.status_code in range(400, 500):
+            raise HTTPError40X
         resp.raise_for_status()
         return resp.json()
 
-    @handle_requests_error
+    @WikipediaBreaker.handle_requests_error
     def get_title_in_language(self, title, source_lang, dest_lang):
         url = self.API_PHP_BASE_PATTERN.format(lang=source_lang)
         resp = self.session.get(
@@ -60,6 +85,8 @@ class WikipediaSession:
             },
             timeout=self.timeout,
         )
+        if resp.status_code in range(400, 500):
+            raise HTTPError40X
         resp.raise_for_status()
         resp_data = resp.json()
         resp_pages = resp_data.get("query", {}).get("pages", [])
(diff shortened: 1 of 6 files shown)


idunn/blocks/wikipedia.py
```diff
@@ -3,11 +3,14 @@
 from apistar import validators
 from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
+import pybreaker
+
+wiki_breaker = pybreaker.CircuitBreaker(fail_max=15, reset_timeout=120)
```
antoine-de: this should be in the configuration (and I think those values are quite low)

idunn/blocks/wikipedia.py
```diff
@@ -3,11 +3,14 @@
...
 
 def handle_requests_error(f):
     def wrapper(*args, **kwargs):
         try:
-            return f(*args, **kwargs)
+            return wiki_breaker.call(f,*args, **kwargs)
```
antoine-de: why don't you catch the circuit breaker error here ?

tests/test_api_circuit_breaker.py
```diff
@@ -0,0 +1,113 @@
...
+    for i in range(20):
+        response = client.get(
+            url=f'http://localhost/v1/pois/{fake_all_blocks}?lang=es',
+        )
+    assert wiki_breaker.fail_counter == len(mock_wiki_requests.calls)
+
+    assert response.status_code == 200
+    resp = response.json()
+
+    assert resp == {
```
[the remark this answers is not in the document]
antoine-de: hum so you should add an explicit assert on the wikipedia block

tests/test_api_circuit_breaker.py
```diff
@@ -0,0 +1,113 @@
...
+    calls).
+    """
+
+    client = TestClient(app)
+    wiki_breaker.close()
+    for i in range(20):
+        response = client.get(
+            url=f'http://localhost/v1/pois/{fake_all_blocks}?lang=es',
+        )
+    assert wiki_breaker.fail_counter == len(mock_wiki_requests.calls)
```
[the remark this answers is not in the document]
antoine-de: hum shouldn't you assert that mock_wiki requests.calls == 15 ?

and you update the configuration for the tests, we can't have the same configuration as the production

idunn/blocks/wikipedia.py
antoine-de: I don't understand the kind of error you catch with the circuit breaker.
we don't want to activate it for 404 or 400.

I think we want it only for 500 and timeout

antoine-de:
if possible it would be nice to add some stats about the circuit breaker in the `/status`

idunn/utils/default_settings.yaml
```diff
@@ -5,6 +5,6 @@ WIKI_USER_AGENT: 'Idunn' # Used in requests to external wiki* APIs
 
 DEFAULT_LANGUAGE: 'en' # Fallback when no 'lang' in request
 
-CIRCUIT_TIMEOUT: 120
```
antoine-de: TIMEOUT seems way too generic.

I think we even need to be more specific and put `WIKI_API_CIRCUIT_TIMEOUT` ?

tests/test_api_circuit_breaker.py
```diff
@@ -50,6 +56,10 @@ def test_circuit_breaker_500(fake_all_blocks, breaker_test):
         resp = response.json()
         assert all(b['type'] != "wikipedia" for b in resp['blocks'])
 
+        """
+        After the timeout the circuit should be half-open.
+        So one more external call should be done.
+        """
         sleep(3)
```
antoine-de: if the reset timeout is 1, why wait 3 seconds ?

idunn/blocks/wikipedia.py
```diff
@@ -3,18 +3,44 @@
...
+    @classmethod
+    def get_breaker(cls):
+        if cls._breaker is None:
+            cls.init_breaker()
+        return cls._breaker
+
+    @classmethod
+    def handle_requests_error(cls, f):
+        def wrapper(*args, **kwargs):
+            if cls._breaker is None:
```
antoine-de: why don't you use `get_breaker`

idunn/blocks/wikipedia.py
```diff
@@ -57,6 +85,8 @@ def get_title_in_language(self, title, source_lang, dest_lang):
             },
             timeout=self.timeout,
         )
+        if resp.status_code in range(400, 500):
```
antoine-de: isn't this too inefficient?  won't this generate 100 integer at each check ?

tests/test_api_circuit_breaker.py
```diff
@@ -0,0 +1,101 @@
...
+
+@pytest.fixture(scope="module", autouse=True)
+def breaker_test():
+    """
+    We define here settings specific to tests
+    for the circuit_breaker in order to avoid
+    any waste of time with real timeout and
+    failmax
+    """
+    WikipediaBreaker.init_breaker()
```
antoine-de: useless method no ?

tests/test_api_circuit_breaker.py
```diff
@@ -0,0 +1,101 @@
...
+    Test that all possible blocks are correct even
+    if the circuit is open.
+    """
+    client = TestClient(app)
+    with responses.RequestsMock() as rsps:
+        rsps.add('GET',
+             re.compile('^https://.*\.wikipedia.org/'),
+             status=500)
+        """
+        We mock 40 calls to wikipedia while the max number
```
antoine-de: the comment is not syncrhonized anymore

tests/test_api_circuit_breaker.py
```diff
@@ -0,0 +1,101 @@
...
+        of failure is 30, so we test that after 30 calls
+        the circuit is open (i.e there are no more than 30
+        calls sent).
+        """
+        breaker_test.close()
```
antoine-de: by curiosity, how could the circuit be open there ? since we don't want side effects between test I don't feel  safe having this kind of stuff here

tests/test_api_circuit_breaker.py
jbgriesner: just to be sure^^

idunn/utils/default_settings.yaml

idunn/blocks/wikipedia.py
(I implemented the method later this is why)

tests/test_api_circuit_breaker.py
jbgriesner: not really: its purpose is to instantiate lazily the breaker to avoid circular dependencies by wrapping the 'import settings'.
Any better way to access the settings ?

idunn/blocks/wikipedia.py
jbgriesner: indeed, done

tests/test_api_circuit_breaker.py
jbgriesner: done

tests/test_api_circuit_breaker.py
jbgriesner: You are right: in the current tests order it cannot be open.
However I thought safer to make each test independent of the order: if later we add a test with the breaker before the first one ('test_circuit_breaker_500') it could let the circuit open and make the successful 'test_circuit_breaker_500' failed (which is exactly what happens to me ^^).
So I just close it at the beginning of each test.

antoine-de: hum i'm not convinced, it's easy to slip by.

maybe close it in the fixture (and make it a yield fixture)

jbgriesner: done

PR #16, opened 2018-07-11
Implement opening_hours "days", other blocks improvements

amatissart:
* Implement detailed "days" in "opening_hours"
* Work around multiple bugs in humanized_opening_hours (waiting for the next release that should solve these issues)
* Do not return "accessibility" block when all values are missing / unknown
* Improve "internet_access" block (using "internet_access" tag from osm)
* Use a session fixture in tests, to mock all external calls to wikipedia by default

### DIFF
--- a/idunn/blocks/opening_hour.py
+++ b/idunn/blocks/opening_hour.py
@@ -1,15 +1,12 @@
-import re
 import logging
-import datetime
-from datetime import datetime
-from pytz import country_timezones, timezone, UTC
-
-from apistar import validators
-from .base import BaseBlock
-
-import humanized_opening_hours as hoh
-from humanized_opening_hours.exceptions import ParseError, SpanOverMidnight
+from datetime import datetime, timedelta, date
+from pytz import timezone, UTC
+from apistar import validators, types
 from tzwhere import tzwhere
+from humanized_opening_hours.exceptions import ParseError, SpanOverMidnight
+
+from idunn.utils.opening_hours_lib import hoh
+from .base import BaseBlock
 
 """
 We load the tz structure once when Idunn starts since it's a time consuming step
@@ -40,6 +37,17 @@ def get_tz(es_poi):
             return timezone(tz.tzNameAt(lat, lon))
     return None
 
+
+class OpeningHoursType(types.Type):
+    beginning = validators.String()
+    end = validators.String()
+
+class DaysType(types.Type):
+    dayofweek = validators.Integer(minimum=1, maximum=7)
+    local_date = validators.Date()
+    status = validators.String(enum=['open', 'closed'])
+    opening_hours = validators.Array(items=OpeningHoursType)
+
 class OpeningHourBlock(BaseBlock):
     BLOCK_TYPE = 'opening_hours'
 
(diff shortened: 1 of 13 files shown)


idunn/blocks/opening_hour.py
```diff
@@ -40,6 +37,17 @@ def get_tz(es_poi):
...
+    end = validators.String()
+
+class DaysType(types.Type):
+    dayofweek = validators.Integer(minimum=1, maximum=7)
+    local_date = validators.Date()
+    status = validators.String(enum=['open', 'closed'])
+    opening_hours = validators.Array(items=OpeningHoursType)
+
 class OpeningHourBlock(BaseBlock):
     BLOCK_TYPE = 'opening_hours'
```
jbgriesner: Minor observation (without link with the PR) but double quote here (ie "opening_hours") is probably better.

idunn/blocks/opening_hour.py
```diff
@@ -74,30 +83,71 @@ def from_es(cls, es_poi, lang):
...
+
+            return cls(
+                status=status,
+                next_transition_datetime=next_transition_datetime,
+                seconds_before_next_transition=time_before_next,
+                is_24_7=is247,
+                raw=oh.sanitized_field,
+                days=cls.get_days(oh, dt=poi_dt)
+            )
+        else:
```
antoine-de: really minor, but I'd prefer an early return

tests/test_opening_hours.py
```diff
@@ -46,12 +45,56 @@ def test_opening_hour_open():
     oh_block = get_moscow_poi("Mo-Su 10:00-22:00")
```
antoine-de: it might be nice to add a opening hour a bit more complex with multiple periods

tests/test_opening_hours.py
```diff
@@ -79,11 +122,91 @@ def test_opening_hour_next_year():
...
+    assert len(oh_block.days) == 7
+    assert all(d.status == 'open' for d in oh_block.days)
+    assert all(d.opening_hours == [dict(
+        beginning='09:00',
+        end='00:00'
+    )] for d in oh_block.days)
+
+
+@freeze_time("2018-06-15T21:00:00+03:00")
+def test_opening_hour_days_cycle():
```
antoine-de: can you add a comment there ?

antoine-de:
there are small conflicts  to be resolved too

idunn/blocks/opening_hour.py
```diff
@@ -58,8 +66,9 @@ def from_es(cls, es_poi, lang):
         is247 = raw == '24/7'
 
         try:
-            clean_oh = hoh.OHParser.sanitize(raw)
-            oh = hoh.OHParser(clean_oh)
+            oh = hoh.OHParser(raw)
+            # Another hack in hoh: apply specific rules in priority
+            oh._tree.tree.children.sort(key=lambda x: len(x[0]))
```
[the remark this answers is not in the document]
amatissart: This sort puts the most specific rules about opening days first.
Without this patch `Mo-Su 10:00-20:00; Th 08:00-22:00` would ignore hours that are defined on Thursdays.

PR #17, opened 2018-07-20
Bypass Wikipedia API

jbgriesner:
Our intern elastic contains enough POI information to bypass the Wikipedia API.
If the request language belongs to the (comma separated string) list ES_WIKI_LANG, then we fetch this elastic, otherwise we request the Wikipedia API as we did before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To run the api you need to do:
 `pipenv install`
 
 and then
-`IDUNN_ES_URL=<url_to_elasticsearch> IDUNN_ES_WIKI_LANG=<comma_separated_languages> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
+`IDUNN_MIMIR_ES=<url_to_elasticsearch> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
 
 you can query the API on 5000:
 `curl localhost:5000/v1/pois/toto?lang=fr`
@@ -36,7 +36,7 @@ The configuration can be given from different ways:
  2. a yaml settings file can be given with an env var IDUNN_CONFIG_FILE
     (the default settings is still loaded and overriden)
  3. specific variable can be overriden with env var. They need to be given like "IDUNN_{var_name}={value}"
-    eg IDUNN_ES_URL=...
+    eg IDUNN_MIMIR_ES=...
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To run the api you need to do:
 `pipenv install`
 
 and then
-`IDUNN_MIMIR_ES=<url_to_elasticsearch> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
+`IDUNN_MIMIR_ES=<url_to_MIMIR_ES> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
 
 you can query the API on 5000:
 `curl localhost:5000/v1/pois/toto?lang=fr`
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To run the api you need to do:
 `pipenv install`
 
 and then
-`IDUNN_ES_URL=<url_to_elasticsearch> pipenv run python app.py`
+`IDUNN_ES_URL=<url_to_elasticsearch> IDUNN_ES_WIKI_LANG=<comma_separated_languages> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
 
 you can query the API on 5000:
 `curl localhost:5000/v1/pois/toto?lang=fr`
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -148,11 +148,8 @@ class WikidataConnector:
             return None
 
         wiki = resp[0]['_source']
-        return (
-            wiki.get("url"),
-            wiki.get("title"),
-            wiki.get("content")
-        )
+
+        return wiki
 
 class WikipediaBlock(BaseBlock):
     BLOCK_TYPE = "wikipedia"
@@ -178,9 +175,9 @@ class WikipediaBlock(BaseBlock):
                     wiki_poi_info = WikidataConnector.get_wiki_info(wikidata_id, lang, wiki_index)
                     if wiki_poi_info is not None:
                         return cls(
-                            url=wiki_poi_info[0],
-                            title=wiki_poi_info[1][0],
-                            description=wiki_poi_info[2],
+                            url=wiki_poi_info.get("url"),
+                            title=wiki_poi_info.get("title")[0],
+                            description=wiki_poi_info.get("content"),
                         )
                     else:
                         """
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -4,6 +4,7 @@ from apistar import validators
 from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
 import pybreaker
+from elasticsearch import Elasticsearch
 
 class HTTPError40X(HTTPError):
     pass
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -24,7 +24,7 @@ To run the api you need to do:
 `pipenv install`
 
 and then
-`IDUNN_ES_URL=<url_to_elasticsearch> pipenv run python app.py`
+`IDUNN_ES_URL=<url_to_elasticsearch> IDUNN_ES_WIKI_LANG=<comma_separated_languages> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
```
antoine-de: in the readme, it might be clearer to only add the the mandatory parameters, so you can bypass the `IDUNN_ES_WIKI_LANG`

idunn/blocks/wikipedia.py
```diff
@@ -100,6 +101,50 @@ def get_title_in_language(self, title, source_lang, dest_lang):
...
+                    }
+                }
+            }
+        ).get('hits', {}).get('hits', [])
+
+        if len(resp) == 0:
+            return None
+
+        wiki = resp[0]['_source']
+        return (wiki.get("url"), wiki.get("title"), wiki.get("content"))
```
antoine-de: I'm not a fan of returning complex tuples, I'd prefer an object

idunn/blocks/wikipedia.py
```diff
@@ -100,6 +101,50 @@ def get_title_in_language(self, title, source_lang, dest_lang):
 
         return None
 
+class Wikidata:
```
antoine-de: the class name seems too generic

idunn/blocks/wikipedia.py
```diff
@@ -111,6 +156,25 @@ class WikipediaBlock(BaseBlock):
...
+            if wiki_index is not None:
+                wiki_poi_info = Wikidata.get_wiki_info(wikidata_id, lang, wiki_index)
+                if wiki_poi_info is not None:
+                    return cls(
+                        url=wiki_poi_info[0],
+                        title=wiki_poi_info[1][0],
+                        description=wiki_poi_info[2],
+                    )
+                else:
+                    return None
```
antoine-de: I would be usefull to explain why we return here

and no need to nest it in a else since the other path returns too

idunn/utils/default_settings.yaml
```diff
@@ -1,10 +1,14 @@
...
+
 WIKI_USER_AGENT: 'Idunn' # Used in requests to external wiki* APIs
 
 DEFAULT_LANGUAGE: 'en' # Fallback when no 'lang' in request
 
 WIKI_API_CIRCUIT_TIMEOUT: 120
 
 WIKI_API_CIRCUIT_MAXFAIL: 50
+
+ES_WIKI_LANG: "de,en,es,fr,it"
```
antoine-de: can you add a comment here ?

tests/conftest.py
```diff
@@ -23,3 +23,50 @@ def es(docker_services):
...
+@pytest.fixture(scope='session')
+def wiki_es(docker_services):
+    """Ensure that ES is up and responsive."""
+    docker_services.start('wiki_es')
+    port = docker_services.wait_for_service("wiki_es", 9200)
+
+    url = f'http://{docker_services.docker_ip}:{port}'
+
+    for c in app.injector.components:
+        if isinstance(c, SettingsComponent):
```
antoine-de: maybe we should add a tests/utils.py with and abstraction to set some settings

tests/conftest.py
```diff
@@ -23,3 +23,50 @@ def es(docker_services):
...
+    return url
+
+
+@pytest.fixture(scope='session')
+def wiki_es_client(wiki_es):
+    return Elasticsearch([wiki_es])
+
+
+@pytest.fixture(scope="session")
+def init_indices(es_client, wiki_es_client):
```
antoine-de: now that we have another es_client it might be nice to rename `es_client` to `mimir_client` ?

tests/docker-compose.yml
```diff
@@ -4,4 +4,9 @@ services:
...
     image: elasticsearch:2-alpine
     ports:
-    - "9200"
\ No newline at end of file
+    - "9200"
+
+  wiki_es:
+    image: elasticsearch:2-alpine
+    ports:
+        - "9201:9200"
```
antoine-de: you shouldn't expose the port as it will makes it impossible to run more than one test instance

tests/test_wiki_ES.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.fixture(scope="session")
+def basket_ball(es_client):
+    """
+    fill elasticsearch with a fake POI of basket ball
+    """
+    return load_poi('basket_ball.json', es_client)
+
+@freeze_time("2018-06-14 8:30:00", tz_offset=2)
+def test_basket_ball(basket_ball, basket_ball_wiki_es):
```
antoine-de: nice test

it would be nice to add a bit more tests though:
* that if ES is not set we still call wikipedia api
* if ES has not the poi we do not call wikipedia
* if we don't have the lang we call wikipedia

idunn/blocks/wikipedia.py
```diff
@@ -100,6 +101,50 @@ def get_title_in_language(self, title, source_lang, dest_lang):
...
+            cls._es_lang = settings['ES_WIKI_LANG'].split(',')
+        if lang in cls._es_lang:
+            return "wikidata_{}".format(lang)
+        return None
+
+    @classmethod
+    def get_wiki_es(cls):
+        if cls._wiki_es is None:
+            from app import settings
+            cls._wiki_es = Elasticsearch(settings['WIKI_ES'])
```
antoine-de: I don't think WIKI_ES should be a mandatory peace of our architecture.

It's quite specific to qwant internal architecture and it would be sad if it made it impossible to use idunn outside qwant

jbgriesner: indeed,
done

tests/test_wiki_ES.py
jbgriesner: nice test ideas,
done

tests/docker-compose.yml
jbgriesner: done

tests/conftest.py
jbgriesner: done

tests/conftest.py
jbgriesner: done, but better can be probably done...

idunn/utils/default_settings.yaml

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py
jbgriesner: done ( "WikidataConnector" )

README.md
jbgriesner: done

tests/test_wiki_ES.py
```diff
@@ -9,27 +9,95 @@
...
+             status=200,
+             json={"test": "test"})
+
+        for i in range(10):
+            response = client.get(
+                url=f'http://localhost/v1/pois/{basket_ball}?lang=fr',
+            )
+
+        assert len(rsps.calls) == 10
+    settings._settings['WIKI_ES'] = wiki_es_ip # put back the correct ip for next tests
```
antoine-de: this won"t be done if the test fail, use a pytest fixture to do this

tests/test_wiki_ES.py
```diff
@@ -9,27 +9,95 @@
...
+    with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
+        rsps.add('GET',
+             re.compile('^https://.*\.wikipedia.org/'),
+             status=200)
+
+        response = client.get(
+                url=f'http://localhost/v1/pois/osm:way:7777777?lang=fr',
+        )
+
+        assert response.status_code == 200
```
antoine-de: can you check more of the response ?

what do you have there ? because mimir does not know it either no ?

idunn/blocks/wikipedia.py

tests/test_wiki_ES.py
```diff
@@ -9,27 +9,95 @@
...
+        for i in range(10):
+            response = client.get(
+                url=f'http://localhost/v1/pois/{basket_ball}?lang=fr',
+            )
+
+        assert len(rsps.calls) == 10
+    settings._settings['WIKI_ES'] = wiki_es_ip # put back the correct ip for next tests
+
+@freeze_time("2018-06-14 8:30:00", tz_offset=2)
+def test_POI_not_in_WIKI_ES(basket_ball, basket_ball_wiki_es):
```
antoine-de: why does it needs the basket_balls features ?

tests/test_wiki_ES.py
jbgriesner: indeed, done

tests/test_wiki_ES.py
jbgriesner: yep, done

tests/test_wiki_ES.py
jbgriesner: was useless, this is fixed

README.md
```diff
@@ -24,7 +24,7 @@ To run the api you need to do:
 `pipenv install`
 
 and then
-`IDUNN_ES_URL=<url_to_elasticsearch> pipenv run python app.py`
+`IDUNN_MIMIR_ES=<url_to_elasticsearch> IDUNN_WIKI_ES=<url_to_WIKI_ES> pipenv run python app.py`
```
antoine-de: I think `<url_to_elasticsearch>` is now a bit vague

PR #18, opened 2018-07-25
Wiki Rate Limiter

jbgriesner:
Because of standard API usage limitations, Idunn cannot make more than 200 requests per second to Wikipedia API.
This PR proposes a rate limiter upon a Redis service.

To activate the limiter you have to define the variable "WIKI_API_REDIS_URL".
If this variable is not set then Idunn doesn't limit its external requests number.

### DIFF
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -5,7 +5,7 @@ from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
 import pybreaker
 from elasticsearch import Elasticsearch
-from redis import ConnectionPool
+from redis import ConnectionPool, ConnectionError, TimeoutError
 
 from redis_rate_limit import RateLimiter, TooManyRequests
 
@@ -15,6 +15,7 @@ class HTTPError40X(HTTPError):
 class WikipediaLimiter:
     _limiter = None
     _redis_url = None
+    _redis_db = None
 
     @classmethod
     def init_limiter(cls):
@@ -31,19 +32,23 @@ class WikipediaLimiter:
             url = cls._redis_url.split(":")[0]
             port = cls._redis_url.split(":")[1]
 
+            cls._redis_db = settings['WIKI_API_REDIS_DB']
+
             cls._limiter = RateLimiter(
                 resource='WikipediaAPI',
                 max_requests=settings['WIKI_API_RL_MAX_CALLS'],
                 expire=settings['WIKI_API_RL_PERIOD'],
-                redis_pool=ConnectionPool(host=url, port=port, db=0)
+                redis_pool=ConnectionPool(host=url, port=port, db=cls._redis_db, socket_timeout=1)
             )
+        else:
+            cls._limiter = False
 
     @classmethod
     def request(cls, f):
         def wrapper(*args, **kwargs):
             if cls._limiter is None:
                 cls.init_limiter()
-            if cls._redis_url is not None:
+            if cls._limiter is not False:
                 """
                 We use the RateLimiter since
                 the redis service url has been provided
@@ -94,6 +99,11 @@ class WikipediaBreaker:
                 logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
             except TooManyRequests:
                 logging.info("Got TooManyRequests{}".format(f.__name__), exc_info=True)
+            except ConnectionError:
+                logging.info("Got redis ConnectionError{}".format(f.__name__), exc_info=True)
+            except TimeoutError:
+                logging.info("Got redis TimeoutError{}".format(f.__name__), exc_info=True)
+
         return wrapper
 
 class WikipediaSession:
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -5,7 +5,58 @@ from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
 import pybreaker
 from elasticsearch import Elasticsearch
+from redis import ConnectionPool
 
+from redis_rate_limit import RateLimiter, TooManyRequests
+
+class HTTPError40X(HTTPError):
+    pass
+
+class WikipediaLimiter:
+    _limiter = None
+    _redis_url = None
+
+    @classmethod
+    def init_limiter(cls):
+        from app import settings
+
+        cls._redis_url = settings['WIKI_API_REDIS_URL']
+
+        if cls._redis_url is not None:
+            """
+            If a redis url has been provided to Idunn,
+            then we use the corresponding redis
+            service in the rate limiter.
+            """
+            url = cls._redis_url.split(":")[0]
+            port = cls._redis_url.split(":")[1]
+
+            cls._limiter = RateLimiter(
+                resource='WikipediaAPI',
+                max_requests=settings['WIKI_API_RL_MAX_CALLS'],
+                expire=settings['WIKI_API_RL_PERIOD'],
+                redis_pool=ConnectionPool(host=url, port=port, db=0)
+            )
+
+    @classmethod
+    def request(cls, f):
+        def wrapper(*args, **kwargs):
+            if cls._limiter is None:
+                cls.init_limiter()
+            if cls._redis_url is not None:
+                """
+                We use the RateLimiter since
+                the redis service url has been provided
+                """
+                with cls._limiter.limit(client="Idunn"):
+                    return f(*args, **kwargs)
+            else:
+                """
+                No redis service has been set, so we
+                bypass the "redis-based" rate limiter
+                """
+                return f(*args, **kwargs)
+        return wrapper
 
 class WikipediaBreaker:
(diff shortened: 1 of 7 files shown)


jbgriesner:
Regarding the Redis resilience I tested locally the scenario of a Redis service interruption with Idunn requests before, during and after the interruption without any problem: Idunn always sends an answer.
Should we do a test for that ?

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,63 @@
...
 
+from redis_rate_limit import RateLimiter, TooManyRequests
+
+class HTTPError40X(HTTPError):
+    pass
+
+class WikipediaLimiter:
+    _limiter = None
+    _redis_url = None
+    _redis_db = None
```
antoine-de: why do you need to store it ?

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,63 @@
...
+
+    @classmethod
+    def init_limiter(cls):
```
antoine-de: minor: it would seems more natural for this to be a lazy getter

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,63 @@
...
+            url = cls._redis_url.split(":")[0]
+            port = cls._redis_url.split(":")[1]
+
+            cls._redis_db = settings['WIKI_API_REDIS_DB']
+
+            cls._limiter = RateLimiter(
+                resource='WikipediaAPI',
+                max_requests=settings['WIKI_API_RL_MAX_CALLS'],
+                expire=settings['WIKI_API_RL_PERIOD'],
+                redis_pool=ConnectionPool(host=url, port=port, db=cls._redis_db, socket_timeout=1)
```
antoine-de: the timeout need to be configurable

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,63 @@
...
+            if cls._limiter is None:
+                cls.init_limiter()
+            if cls._limiter is not False:
+                """
+                We use the RateLimiter since
+                the redis service url has been provided
+                """
+                with cls._limiter.limit(client="Idunn"):
+                    return f(*args, **kwargs)
+            else:
```
antoine-de: minor: you don't need this `else` nesting

idunn/blocks/wikipedia.py
jbgriesner: removed

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py

jbgriesner:
@antoine-de I added the test 'test_rate_limiter_with_redis_interruption'

tests/test_rate_limiter.py
```diff
@@ -1,53 +1,122 @@
...
     """
     settings._settings['WIKI_API_RL_PERIOD'] = 5
     settings._settings['WIKI_API_RL_MAX_CALLS'] = 3
     settings._settings['WIKI_API_REDIS_URL'] = redis
     WikipediaLimiter._limiter = None
-    wiki_limiter = WikipediaLimiter.init_limiter()
     yield
-    settings._settings['WIKI_API_RL_PERIOD'] = 100
-    settings._settings['WIKI_API_RL_MAX_CALLS'] = 1
+    settings._settings['WIKI_API_RL_PERIOD'] = 1
```
antoine-de: don't you want to set it back to the old param values ?
else this will create a test side effect

tests/test_rate_limiter.py
```diff
@@ -1,53 +1,122 @@
...
         )
         """
-        We mock 10 calls to wikipedia while the max number
+        We mock 5 calls to wikipedia while the max number
         of calls is 3, so we test that after 3 calls
         no more calls are done
         """
 
-        for i in range(10):
+        for i in range(5):
```
antoine-de: hum maybe this test would be better is written something like:

```python" 
# for the first calls, we have a wikipedia block
for i in range(3):
    response = client.get(...)
    resp = response.json()
    assert any(b['type'] == "wikipedia" for b in resp['blocks'])

# then we stop calling wikipedia, so there is no block
for i in range(2):
    response = client.get(...)
    resp = response.json()
    assert al(b['type'] != "wikipedia" for b in resp['blocks'])

assert len(rsps.calls) == 3
```

tests/test_rate_limiter.py
```diff
@@ -60,23 +129,129 @@ def test_rate_limiter_with_redis(fake_all_blocks, limiter_test):
...
+def test_rate_limiter_without_redis(louvre_museum):
     """
     Test that Idunn doesn't stop external requests when
     no redis has been set: 10 requests to Idunn should
     generate 10 requests to Wikipedia API
     """
     client = TestClient(app)
-    with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
+
+    WikipediaLimiter._limiter = None
```
antoine-de: why do you need this ?

this should be in a yield fixture

tests/test_rate_limiter.py
```diff
@@ -60,23 +129,129 @@ def test_rate_limiter_with_redis(fake_all_blocks, limiter_test):
...
             )
 
         assert len(rsps.calls) == 10
+
+
+def restart_wiki_redis(docker_services):
+    """
+    Because docker services ports are
+    dynamically chosen when the service starts
+    we have to get the new port of the service.
```
antoine-de: is'nt this beecause we do not have stopped the previous service ?

jbgriesner: Wasn't it stop line 211 ?

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,67 @@
...
+            cls._redis_url = settings['WIKI_API_REDIS_URL']
+
+            if cls._redis_url is not None:
+                """
+                If a redis url has been provided to Idunn,
+                then we use the corresponding redis
+                service in the rate limiter.
+                """
+                url = cls._redis_url.split(":")[0]
+                port = cls._redis_url.split(":")[1]
```
antoine-de: this won't work if no port is defined

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,67 @@
...
+class WikipediaLimiter:
+    _limiter = None
+    _redis_url = None
+
+    @classmethod
+    def get_limiter(cls):
+        if cls._limiter is None:
+            from app import settings
+
+            cls._redis_url = settings['WIKI_API_REDIS_URL']
```
antoine-de: I don't think we need to store this

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,67 @@
...
+            if cls._redis_url is not None:
+                """
+                If a redis url has been provided to Idunn,
+                then we use the corresponding redis
+                service in the rate limiter.
+                """
+                url = cls._redis_url.split(":")[0]
+                port = cls._redis_url.split(":")[1]
+
+                max_calls= settings['WIKI_API_RL_MAX_CALLS']
```
antoine-de: it does not seems to work in real like as when read from the config this is a string and this results in error like
```
  File "/dev/idunn/idunn/blocks/wikipedia.py", line 61, in wrapper
    with limiter.limit(client="Idunn"):
  File "/.local/share/virtualenvs/idunn-OLqrlkmx/lib/python3.6/site-packages/redis_rate_limit/__init__.py", line 69, in __enter__
    self.increment_usage()
  File "/.local/share/virtualenvs/idunn-OLqrlkmx/lib/python3.6/site-packages/redis_rate_limit/__init__.py", line 122, in increment_usage
    if int(current_usage) > self._max_requests:
TypeError: '>' not supported between instances of 'int' and 'str'
```

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py
```diff
@@ -14,35 +14,39 @@ class HTTPError40X(HTTPError):
...
+                    redis_db = settings['WIKI_API_REDIS_DB']
+                    redis_timeout = int(settings['WIKI_REDIS_TIMEOUT'])
+
+                    cls._limiter = RateLimiter(
+                        resource='WikipediaAPI',
+                        max_requests=max_calls,
+                        expire=redis_period,
+                        redis_pool=ConnectionPool(host=ip, port=port, db=redis_db, socket_timeout=redis_timeout)
+                    )
+                except AssertionError:
```
antoine-de: It's quite dangerous to catch AssertionError what if it was the RateLimiter that raised one ?

But I'm sorry my comment was irrelevent as the `ConnectionPool` enfore the use of a port, so anyway the port is mandatory in the url

it's realy minor, but maybe you can only rewrite the url/port ready like:
```python
url, port = cls._redis_url.split(":")
```

jbgriesner: thanks
done

PR #19, opened 2018-07-26
Wiki ES timeout

jbgriesner:
When the Wiki ES service is down we want Idunn continue to deliver the correct information.
In this small patch we catch the 'connectionerror' to avoid the Idunn failure.

+ small test

### DIFF
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -209,7 +209,7 @@ class WikidataConnector:
                 }
             ).get('hits', {}).get('hits', [])
         except ConnectionError:
-            logging.warning("Wiki ES not available: exception raised in {}".format(f.__name__), exc_info=True)
+            logging.warning("Wiki ES not available: connection exception raised", exc_info=True)
             return None
 
         if len(resp) == 0:
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -4,7 +4,7 @@ from apistar import validators
 from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
 import pybreaker
-from redis import ConnectionPool, ConnectionError, TimeoutError
+from redis import ConnectionPool, ConnectionError as RedisConnectionError, TimeoutError
 from elasticsearch import Elasticsearch, ElasticsearchException, RequestError, ConnectionTimeout, ConnectionError
 
 from redis_rate_limit import RateLimiter, TooManyRequests
@@ -101,7 +101,7 @@ class WikipediaBreaker:
                 logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
             except TooManyRequests:
                 logging.info("Got TooManyRequests{}".format(f.__name__), exc_info=True)
-            except ConnectionError:
+            except RedisConnectionError:
                 logging.info("Got redis ConnectionError{}".format(f.__name__), exc_info=True)
             except TimeoutError:
                 logging.info("Got redis TimeoutError{}".format(f.__name__), exc_info=True)
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -198,7 +198,7 @@ class WikidataConnector:
                 cls._wiki_es = Elasticsearch(
                         wiki_es,
                         max_retries=wiki_es_max_retries,
-                        request_timeout=wiki_es_timeout
+                        timeout=wiki_es_timeout
                 )
 
     @classmethod
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -187,11 +187,19 @@ class WikidataConnector:
     def init_wiki_es(cls):
         if cls._wiki_es is None:
             from app import settings
+
             wiki_es = settings.get('WIKI_ES')
+            wiki_es_max_retries = settings.get('WIKI_ES_MAX_RETRIES')
+            wiki_es_timeout = settings.get('WIKI_ES_TIMEOUT')
+
             if wiki_es is None:
                 raise WikiUndefinedException
             else:
-                cls._wiki_es = Elasticsearch(wiki_es)
+                cls._wiki_es = Elasticsearch(
+                        wiki_es,
+                        max_retries=wiki_es_max_retries,
+                        request_timeout=wiki_es_timeout
+                )
 
     @classmethod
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -34,11 +34,26 @@ def wiki_es(docker_services):
...
+@pytest.fixture(scope='session')
+def wiki_es_ko(docker_services):
+    docker_services.start('wiki_es')
+    port = docker_services.wait_for_service("wiki_es", 9200)
+
+    url = "1.1.1.1:1234"
+
+    settings._settings['WIKI_ES'] = "1.1.1.1:1234"
+    settings._settings['ES_WIKI_LANG'] = 'fr'
+    return url
```
amatissart: this fixture should use a different scope, and revert the settings after each test.
Also, I would use a fake domain (such as "something.invalid", .test is reserved for such usecases) instead of a real address.

tests/conftest.py
```diff
@@ -34,11 +34,26 @@ def wiki_es(docker_services):
...
+
+@pytest.fixture(scope='session')
+def wiki_client_ko(wiki_es):
+    return Elasticsearch([wiki_es])
```
amatissart: I think that `wiki_es_ko` shoud be used here instead.

tests/test_wiki_ES.py
```diff
@@ -42,6 +42,72 @@ def undefine_wiki_es():
...
+                    }
+                ]
+            }
+        ]
+
+        """
+        We check that there is no Wikipedia block
+        in the answer
+        """
+        assert all(b['type'] != "wikipedia" for b in resp['blocks'])
```
amatissart: This check is incorrect. If the wikipedia block is present, it would be nested inside the "information" block.

tests/conftest.py
jbgriesner: done

tests/conftest.py
jbgriesner: indeed^^ done

tests/test_wiki_ES.py
jbgriesner: done

tests/conftest.py
```diff
@@ -22,23 +22,52 @@ def mimir_es(docker_services):
...
 def wiki_client(wiki_es):
     return Elasticsearch([wiki_es])
 
+
+@pytest.fixture(scope='session')
+def wiki_es_ko(docker_services):
+    docker_services.start('wiki_es')
+    port = docker_services.wait_for_service("wiki_es", 9200)
+
+    temp_wiki_es = settings._settings['WIKI_ES']
```
antoine-de: since it's something that seems to be done often, I think it's worth it do make it more generic.

maybe something like (not tested, just speudo code):

```python
@pytest.fixture(scope='session')
def wiki_es_ko(docker_services):
    ...
    with utils.override_settings({'WIKI_ES': url, 'ES_WIKI_LANG': 'fr'}):
        yield
```

and in some test/utils.py file
```python
from contextlib import contextmanager
from copy import deepcopy
from app import settings

@contextmanager
def override_settings(overrides):
    old_settings = deepcopy(settings._settings)
    settings._settings.update(overrides)
    yield 
    settings._settings = old_settings
 ```

Comment (jbgriesner) [reply tests/conftest.py]: Nice,
done

File: tests/conftest.py [hunk 5]
```diff
@@ -22,23 +22,52 @@ def mimir_es(docker_services):
...
 
-    url = f'http://{docker_services.docker_ip}:{port}'
+    temp_wiki_es = settings._settings['WIKI_ES']
+    temp_es_wiki_lang = settings._settings['ES_WIKI_LANG']
 
+    url = f'http://{docker_services.docker_ip}:{port}'
     settings._settings['WIKI_ES'] = url
     settings._settings['ES_WIKI_LANG'] = 'fr'
+
     return url
```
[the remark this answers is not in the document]
jbgriesner: ok, done

tests/conftest.py
```diff
@@ -22,23 +22,52 @@ def mimir_es(docker_services):
...
+    port = docker_services.wait_for_service("wiki_es", 9200)
+
+    temp_wiki_es = settings._settings['WIKI_ES']
+    temp_es_wiki_lang = settings._settings['ES_WIKI_LANG']
+
+    url = "something.invalid:1234"
+    settings._settings['WIKI_ES'] = url
+    settings._settings['ES_WIKI_LANG'] = 'fr'
+
+    return url
```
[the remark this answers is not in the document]
jbgriesner: done

tests/conftest.py
```diff
@@ -34,11 +34,26 @@ def wiki_es(docker_services):
...
 @pytest.fixture(scope='session')
 def wiki_client(wiki_es):
     return Elasticsearch([wiki_es])
 
+@pytest.fixture(scope='session')
+def wiki_es_ko(docker_services):
+    docker_services.start('wiki_es')
+    port = docker_services.wait_for_service("wiki_es", 9200)
+
+    url = "1.1.1.1:1234"
```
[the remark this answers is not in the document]
jbgriesner: indeed, but it has been changed in commit https://github.com/QwantResearch/idunn/pull/19/commits/5596ea36fb49a401e517a3dddade83b072178379

idunn/blocks/wikipedia.py
```diff
@@ -4,8 +4,8 @@
 from .base import BaseBlock, BlocksValidator
 from requests.exceptions import HTTPError, RequestException, Timeout
 import pybreaker
-from elasticsearch import Elasticsearch
-from redis import ConnectionPool, ConnectionError, TimeoutError
+from redis import ConnectionPool, ConnectionError as RedisConnectionError, TimeoutError
+from elasticsearch import Elasticsearch, ElasticsearchException, RequestError, ConnectionTimeout, ConnectionError
```
amatissart: Most of these imports are not used

idunn/blocks/wikipedia.py
```diff
@@ -197,16 +197,20 @@ def init_wiki_es(cls):
...
+                    "filter": {
+                        "term": {
+                            "wikibase_item": wikidata_id
+                        }
                     }
                 }
-            }
-        ).get('hits', {}).get('hits', [])
+            ).get('hits', {}).get('hits', [])
+        except ConnectionError:
```
amatissart: I think we should also override Elasticsearch `timeout` and `max_retries` params. 10s and 3 retries is too much here.

jbgriesner: Good idea, done

idunn/blocks/wikipedia.py

PR #20, opened 2018-07-26
tactile_paving removed

jbgriesner:
This PR adapts the accessibility block to the following specifications:

* if wheelchair and toilets:wheelchair are both undefined : no accessibility block
* if wheelchair and toilets:wheelchair are both defined with unknown values: no accessibility block

Simple tests (from the specifications) come along the changes.

### DIFF
--- a/idunn/blocks/services_and_information.py
+++ b/idunn/blocks/services_and_information.py
@@ -24,12 +24,6 @@ class AccessibilityBlock(BaseBlock):
         raw_wheelchair = properties.get("wheelchair")
         raw_toilets_wheelchair = properties.get("toilets:wheelchair")
 
-        if all(
-            s is None
-            for s in (raw_wheelchair, raw_toilets_wheelchair)
-        ):
-            return None
-
         if raw_wheelchair in ("yes", "designated"):
             wheelchair = cls.STATUS_OK
         elif raw_wheelchair == "limited":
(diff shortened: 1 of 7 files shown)


idunn/blocks/services_and_information.py
```diff
@@ -23,9 +22,14 @@ def from_es(cls, es_poi, lang):
...
 
         raw_wheelchair = properties.get("wheelchair")
-        raw_tactile_paving = properties.get("tactile_paving")
         raw_toilets_wheelchair = properties.get("toilets:wheelchair")
 
+        if all(
+            s is None
+            for s in (raw_wheelchair, raw_toilets_wheelchair)
+        ):
+            return None
```
amatissart: The check is redundant and can be removed.

tests/test_accessibility.py
```diff
@@ -25,9 +25,68 @@ def test_accessibility_unknown():
...
+    assert resp['name'] == "Boulangerie Patisserie Peron"
+    assert resp['local_name'] == "Boulangerie Patisserie Peron"
+    assert resp['class_name'] == 'bakery'
+    assert resp['subclass_name'] == 'bakery'
+    assert resp['blocks'] == []
+
+def test_wheelchair(cinema_multiplexe):
+    """
+    Test that when wheelchair is defined but not toilets_wheelchair
+    we still have a correct block
```
amatissart: this description does not match the test, where both values are "yes"

idunn/blocks/services_and_information.py
jbgriesner: your are right: removed

tests/test_accessibility.py
jbgriesner: Indeed: it has been fixed

PR #22, opened 2018-08-11
Cache external requests

jbgriesner:
This PR proposes a simple cache implementation of the calls to external services.
It uses the pyMemoize library.
Simple tests come along the PR.


idunn/blocks/wikipedia.py
```diff
@@ -108,6 +113,46 @@ def wrapper(*args, **kwargs):
 
         return wrapper
 
+class WikipediaCache:
+    _cache = None
```
amatissart: we need to set the expire timeout value somewhere, and make it configurable via settings.
Persist the wikipedia values for 24 hours is probably a good default.

amatissart: I think we should add a setting to set the expire timeout of values that are stored in Redis.
24h is probably a good default for wikipedia values.

jbgriesner: done.

amatissart:
I am still unsure that using `PyMemoize` is a good idea here. It seems the redis integration implements its own expiration mechanism (https://github.com/mikeboers/PyMemoize/blob/v1.0.2/memoize/core.py#L49). This means old keys are kept permanently in Redis.
It would be more straightforward to use the redis SET options (https://github.com/mikeboers/PyMemoize/blob/v1.0.2/memoize/core.py#L49) that handle cleanup automatically.

jbgriesner:
done.

idunn/blocks/wikipedia.py
```diff
@@ -118,6 +148,69 @@ def wrapper(*args, **kwargs):
...
+            """
+            If the redis is up we use the cache,
+            otherwise we bypass it
+            """
+            if cls._connection is not False:
+                """
+                The key is the function name (prefix)
+                along with the args.
+                """
+                args_list = [arg for ak in zip(args, kwargs.values()) for arg in ak]
```
amatissart: This seems broken to me.
How zip is supposed to work here ? And `self` should be removed from args.  
We probably don't need this complexity. Why not passing a complete key to `cache_it` that could be used in redis directly ?

tests/test_cache.py
```diff
@@ -0,0 +1,130 @@
...
+            Without the cache the request would fail
+            in the "get_wiki_info()" method
+            """
+            for i in range(10):
+                response = client.get(
+                    url=f'http://localhost/v1/pois/{basket_ball}?lang=fr',
+                )
+
+            assert len(rsps.calls) == 0 # Wikipedia API has never been called
+            assert any(b['type'] == "wikipedia" for b in resp['blocks'][2].get('blocks')) # we still have the wikipedia block
```
amatissart: This assert should be in the loop, and use the new `response` instead of the previous `resp` (that does not use the cache :wink:)

jbgriesner:
done.

idunn/blocks/wikipedia.py
```diff
@@ -118,6 +129,64 @@ def wrapper(*args, **kwargs):
...
+
+        def with_cache(*args, **kwargs):
+            """
+            If the redis is up we use the cache,
+            otherwise we bypass it
+            """
+            if cls._connection is not False:
+
+                if cls._connection.exists(key):
+                    result = json.loads(cls._connection.get(key).decode('utf-8'))
```
amatissart: Here there are 2 requests to redis: `EXISTS` and `GET`.  
A single `.get` should be enough and will return `None` if the key does not exist.

jbgriesner:
done.

PR #24, opened 2018-08-23
Resilience to Redis fail

jbgriesner:
Just add a catch to the most general redis exception raised by a lack of disk space.


antoine-de:
This does not work as `.limiter` does not exists.

I also think that changing `WikipediaLimiter` class variables should be the responsability of `WikipediaLimiter` not `WikipediaBreaker`

And this the limiter is not working we should not call wikipedia at all, else we could flood them.

PR #26, opened 2018-09-04
Disk space - redis

jbgriesner:
Based on the previous PR (https://github.com/QwantResearch/idunn/pull/24) this PR changes the behavior of Idunn when the disk is full and stops returning a wikipedia block in order to not flood the wiki api.


jbgriesner:
done.

PR #28, opened 2018-09-12
Prometheus first metrics

jbgriesner:
This PR adds the `/v1/metrics` endpoint.
It is based on the [Prometheus Python Client](https://github.com/prometheus/client_python) and also on the [apistar_prometheus](https://github.com/Bogdanp/apistar_prometheus)

The endpoint is used by Prometheus to plot some metrics.
By default [apistar_prometheus](https://github.com/Bogdanp/apistar_prometheus) comes along with 3 metrics:
- `REQUEST_DURATION` : Time spent processing a request
- `REQUEST_COUNT` : Request count by method, handler and response code
- `REQUESTS_INPROGRESS` : Requests in progress by method and handler

Below is an example graph of the total number of requests (`REQUEST_COUNT` metric) received by the corresponding handler:

selection_003


tests/test_metrics.py
```diff
@@ -0,0 +1,13 @@
+from unittest.mock import patch
```
antoine-de: why do you need this ?

jbgriesner: my bad, 
removed

PR #29, opened 2018-09-18
Add more logging facility for idunn

### LINKED ISSUE #21: Configure logging
The logger is not initialized in idunn, so the default log level, `warn` is used.

We should add a way to initlialize the logger.

I think we should also create a logger by module to have a fine grained control over it

antoine-de:
The main thing is that we are now able to output json logs (for easier
ELK integration).

To enable it, the `IDUNN_LOG_JSON` must be set to 1.

`LOG_FORMAT` can be used to defined the logged fields / format
`LOG_LEVEL_BY_MODULE` can be set to define the log level for each module

Gunicorn now also output its log as json (and this is not configurable,
it's set in the Dockerfile)

should close #21

### DIFF
--- a/gunicorn_logging.conf
+++ b/gunicorn_logging.conf
@@ -1,0 +1,20 @@
+[loggers]
+keys=root
+
+[handlers]
+keys=console
+
+[formatters]
+keys=json
+
+[logger_root]
+level=INFO
+handlers=console
+
+[handler_console]
+class=StreamHandler
+formatter=json
+args=(sys.stdout, )
+
+[formatter_json]
+class=pythonjsonlogger.jsonlogger.JsonFormatter
--- a/gunicorn_logging.conf
+++ b/gunicorn_logging.conf
@@ -1,5 +1,5 @@
 [loggers]
-keys=root
+keys=root, gunicorn.error, gunicorn.access
 
 [handlers]
 keys=console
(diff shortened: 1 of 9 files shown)


gunicorn_logging.conf
```diff
@@ -0,0 +1,20 @@
+[loggers]
+keys=root
```
amatissart: Maybe add `gunicorn.error` and `gunicorn.access` here ?

gunicorn_logging.conf
```diff
@@ -0,0 +1,20 @@
...
+level=INFO
+handlers=console
+
+[handler_console]
+class=StreamHandler
+formatter=json
+args=(sys.stdout, )
+
+[formatter_json]
+class=pythonjsonlogger.jsonlogger.JsonFormatter
```
amatissart: This formatter needs a `format` to output more fields

idunn/utils/logging.py
```diff
@@ -0,0 +1,38 @@
...
+    else:
+        logHandler.setFormatter(logging.Formatter(log_format))
+    
+    # we set this handler to the main logger
+    logging.getLogger().handlers = [logHandler]
+
+class LogErrorHook:
+    def on_response(self, response: http.Response, exc: Exception):
+        if exc is not None:
+            logging.getLogger('idunn.error').exception("handler returned an error")
```
amatissart: Here `exc` is necessarily a subclass of `HTTPException`, that is directly handled by apistar to build a response.
As a result any 404 (for example) is logged here as an ERROR, which may be not be relevant.

Should we keep the `on_error` hook only ?

PR #30, opened 2018-09-18
More metrics

jbgriesner:
Basically this adds the 4 following Prometheus metrics:
- `WIKI_REQUEST_DURATION` (histogram)
- `WIKI_BREAKER_STATUS` (gauge)
- `WIKI_BREAKER_EXCEPTIONS_COUNT` (counter)
- `WIKI_RATE_LIMITER_MAX_COUNT` (counter)

These metrics are available on the `/v1/metrics` endpoint.

Below is an illustration of a raw output of this endpoint:

selection_002

selection_001

### DIFF
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -113,7 +113,7 @@ class WikipediaBreaker:
             try:
                 return WikipediaLimiter.request(breaker(f))(*args, **kwargs)
             except pybreaker.CircuitBreakerError:
-                prometheus.breaker_error()
+                prometheus.exception("CircuitBreakerError")
                 logging.error("Got CircuitBreakerError in {}".format(f.__name__), exc_info=True)
             except HTTPError:
                 prometheus.exception("HTTPError")
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -5,11 +5,11 @@ import pybreaker
 
 from apistar import validators
 from .base import BaseBlock, BlocksValidator
-from idunn.utils.prometheus import wiki_request_duration, breaker_error, exception, limiter_exception
 from requests.exceptions import HTTPError, RequestException, Timeout
 from redis import Redis, ConnectionPool, ConnectionError as RedisConnectionError, TimeoutError, RedisError
 from elasticsearch import Elasticsearch, ConnectionError
 from redis_rate_limit import RateLimiter, TooManyRequests
+from idunn.utils import prometheus
 
 GET_WIKI_INFO = "get_wiki_info"
 GET_TITLE_IN_LANGUAGE = "get_title_in_language"
@@ -113,25 +113,25 @@ class WikipediaBreaker:
             try:
                 return WikipediaLimiter.request(breaker(f))(*args, **kwargs)
             except pybreaker.CircuitBreakerError:
-                breaker_error()
+                prometheus.breaker_error()
                 logging.error("Got CircuitBreakerError in {}".format(f.__name__), exc_info=True)
             except HTTPError:
-                exception("HTTPError")
+                prometheus.exception("HTTPError")
                 logging.warning("Got HTTP error in {}".format(f.__name__), exc_info=True)
             except Timeout:
-                exception("RequestsTimeout")
+                prometheus.exception("RequestsTimeout")
                 logging.warning("External API timed out in {}".format(f.__name__), exc_info=True)
             except RequestException:
-                exception("RequestException")
+                prometheus.exception("RequestException")
                 logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
             except TooManyRequests:
-                limiter_exception()
+                prometheus.exception("TooManyRequests")
                 logging.warning("Got TooManyRequests{}".format(f.__name__), exc_info=True)
             except RedisConnectionError:
-                exception("RedisConnectionError")
+                prometheus.exception("RedisConnectionError")
                 logging.warning("Got redis ConnectionError{}".format(f.__name__), exc_info=True)
             except TimeoutError:
-                exception("RedisTimeoutError")
+                prometheus.exception("RedisTimeoutError")
                 logging.warning("Got redis TimeoutError{}".format(f.__name__), exc_info=True)
         return wrapped_f
 
@@ -215,7 +215,7 @@ class WikipediaSession:
             base_url=self.API_V1_BASE_PATTERN.format(lang=lang), title=title
         )
 
-        with wiki_request_duration("wiki_api", "get_summary"):
+        with prometheus.wiki_request_duration("wiki_api", "get_summary"):
             resp = self.session.get(
                 url=url,
                 params={"redirect": True},
@@ -231,7 +231,7 @@ class WikipediaSession:
     def get_title_in_language(self, title, source_lang, dest_lang):
         url = self.API_PHP_BASE_PATTERN.format(lang=source_lang)
 
-        with wiki_request_duration("wiki_api", "get_title"):
+        with prometheus.wiki_request_duration("wiki_api", "get_title"):
             resp = self.session.get(
                 url=url,
                 params={
@@ -302,7 +302,7 @@ class WikidataConnector:
         cls.init_wiki_es()
 
         try:
-            with wiki_request_duration("wiki_es", "get_wiki_info"):
+            with prometheus.wiki_request_duration("wiki_es", "get_wiki_info"):
                 resp = cls._wiki_es.search(
                     index=wiki_index,
                     body={
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -85,12 +85,10 @@ class LogListener(pybreaker.CircuitBreakerListener):
 
     def state_change(self, cb, old_state, new_state):
         msg = "State Change: CB: {0}, From: {1} to New State: {2}".format(cb.name, old_state, new_state)
-        WikipediaBreaker.update_status()
         logging.warning(msg)
 
 class WikipediaBreaker:
     _breaker = None
-    _status = 'closed'
     _prom_tracker = None
 
     @classmethod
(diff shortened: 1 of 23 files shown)


idunn/utils/prometheus.py
```diff
@@ -0,0 +1,69 @@
...
+
+class PrometheusTracker:
+    """ The logic of the Prometheus metrics is defined in this component """
+
+    _tracker = None
+
+    class _Prometheus:
+        """ Singleton that catch some primitives on the Prometheus metrics """
+
+        def request_start(self, handler):
```
antoine-de: I think the split `request_start`/`request_end` method can lead to many errors, a context manager would be better

```python

    @contextlib.contextmanager
    def log_request_duration(self, handler):
        with WIKI_REQUEST_DURATION.labels(handler).time():
            yield
```

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,69 @@
...
+
+        def breaker_exception(self):
+            WIKI_BREAKER_EXCEPTIONS_COUNT.inc()
+
+        def limiter_exception(self):
+            WIKI_RATE_LIMITER_MAX_COUNT.inc()
+
+        def update_status(self, old_status, new_status):
+            diff = BREAKER_STATUS.get(old_status) - BREAKER_STATUS.get(new_status)
+            if (lambda x: (1, -1)[x < 0])(diff) == 1:
```
antoine-de: I'm not sure, but I think this can be simplified: `if diff > 0:`

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,69 @@
...
+                for i in range(0, diff):
+                    WIKI_BREAKER_STATUS.inc()
+            else:
+                for i in range(0, abs(diff)):
+                    WIKI_BREAKER_STATUS.dec()
+
+    @classmethod
+    def get_tracker(cls):
```
antoine-de: can't you put this a a free method and remove the `_Prometheus` class

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+WIKI_BREAKER_STATUS = Gauge(
+    "breaker_status",
+    "Breaker status, open (=2) or half-open (=1) or close (=0).",
+    []
+)
+
+WIKI_BREAKER_EXCEPTIONS_COUNT = Counter(
+    "breaker_exceptions_count",
```
antoine-de: I think all prometheus metrics should be prefixed by `idunn_`

idunn/blocks/wikipedia.py
```diff
@@ -190,11 +202,16 @@ def with_cache(*args, **kwargs):
...
     _session = None
+    _prom_tracker = None
     timeout = 1. # seconds
 
     API_V1_BASE_PATTERN = "https://{lang}.wikipedia.org/api/rest_v1"
     API_PHP_BASE_PATTERN = "https://{lang}.wikipedia.org/w/api.php"
 
+    def __init__(self):
+        if self._prom_tracker is None:
+            self._prom_tracker = PrometheusTracker.get_tracker()
```
antoine-de: you don't need to store the tracker since it's a singleton

idunn/blocks/wikipedia.py
```diff
@@ -122,13 +134,13 @@ def wrapper(*args, **kwargs):
             except RequestException:
                 logging.error("Got Request exception in {}".format(f.__name__), exc_info=True)
             except TooManyRequests:
+                cls._prom_tracker.limiter_exception()
                 logging.warning("Got TooManyRequests{}".format(f.__name__), exc_info=True)
             except RedisConnectionError:
                 logging.warning("Got redis ConnectionError{}".format(f.__name__), exc_info=True)
             except TimeoutError:
```
antoine-de: I think we need ton counter for each of these exception non ? (maybe just a handler on a generic exception metric ?)

idunn/utils/prometheus.py
jbgriesner: indeed :+1:

idunn/utils/prometheus.py
jbgriesner: done

idunn/utils/prometheus.py
jbgriesner: it can^^

idunn/utils/prometheus.py
jbgriesner: done

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py
```diff
@@ -218,18 +237,21 @@ def get_summary(self, title, lang):
...
-                "prop": "langlinks",
-                "titles": title,
-                "lllang": dest_lang,
-                "formatversion": 2,
-                "format": "json",
-            },
-            timeout=self.timeout,
-        )
+
+        with self._prom_tracker.log_request_duration("wiki_api"):
```
antoine-de: shouldn't we have different handler for the title than the sumary ?

idunn/utils/default_settings.yaml
```diff
@@ -24,3 +24,4 @@ WIKI_CACHE_TIMEOUT: 86400 # seconds
 LOG_LEVEL_BY_MODULE: '{"": "info", "elasticsearch": "warning"}' # json config to set, for each module a log level
 LOG_FORMAT: '[%(asctime)s] [%(levelname)5s] [%(process)5s] [%(name)10s] %(message)s' # logging format. if the log are json, it list the default fields
 LOG_JSON: False  # To get flat logs or json logs
+PROMETHEUS_MULTIPROC: False
```
antoine-de: there should be a comment here

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,50 @@
...
+    "idunn_limiter_too_many_requests_count",
+    "Number of times TooManyRequests exception has been raised.",
+    []
+)
+
+class PrometheusTracker:
+    """ Just catch some primitives on the Prometheus metrics """
+
+    @contextlib.contextmanager
+    def log_request_duration(self, handler):
```
antoine-de: if the metric is only for wikipedia, there should be wikipedia in the name of the function

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,50 @@
...
+        with IDUNN_WIKI_REQUEST_DURATION.labels(handler).time():
+            yield
+
+    def breaker_error(self):
+        IDUNN_WIKI_BREAKER_ERRORS_COUNT.inc()
+
+    def breaker_exception(self, exception_type):
+        IDUNN_WIKI_BREAKER_EXCEPTIONS_COUNT.labels(exception_type).inc()
```
antoine-de: I don't think it's breaker centric

idunn/blocks/wikipedia.py
```diff
@@ -265,6 +289,8 @@ def get_wiki_index(cls, lang):
 
     @classmethod
     def init_wiki_es(cls):
+        if cls._prom_tracker is None:
```
antoine-de: I don't think building a `PrometheusTracker` and keeping it as a class variable is useful.

I tihnk we can either:

* remove the PrometheusTracker class, and make all the function module function
we will then use it as `prometheus.exception("RequestException")`

* Instantiate a PrometheusTracker in the prometheus module
we will use it as `prometheus.tracker.exception("RequestException")`

I'd vote for the first solution, but the second one is fine for me too

idunn/blocks/wikipedia.py
jbgriesner: Since both are requesting the wiki api, I thought a single handler was ok for both of them.
No ?

idunn/blocks/wikipedia.py
jbgriesner: Ok I go for the first one then

idunn/utils/default_settings.yaml

idunn/utils/prometheus.py
jbgriesner: done

idunn/blocks/wikipedia.py
antoine-de: it's strange to aggregate the metrics on those calls (than ban be quite different)

jbgriesner: yup, done

idunn/utils/prometheus.py
jbgriesner: done

idunn/blocks/wikipedia.py
```diff
@@ -5,7 +5,7 @@
 
 from apistar import validators
 from .base import BaseBlock, BlocksValidator
-from idunn.utils.prometheus import PrometheusTracker
+from idunn.utils.prometheus import wiki_request_duration, breaker_error, exception, limiter_exception
```
antoine-de: I think it's better to only import the module and use it:
`prometheus.wiki_request_duration`

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,47 @@
...
+    ["target", "handler"],
+)
+
+IDUNN_WIKI_EXCEPTIONS_COUNT = Counter(
+    "idunn_exceptions_count",
+    "Number of exceptions caught in Idunn.",
+    ["exception_type"]
+)
+
+IDUNN_WIKI_BREAKER_ERRORS_COUNT = Counter(
```
antoine-de: I'm not sure we want custom counter for the breaker and rate_limiter

idunn/utils/prometheus.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+IDUNN_WIKI_REQUEST_DURATION = Histogram(
+    "idunn_wiki_request_duration_seconds",
+    "Time spent processing a Wiki request.",
+    ["target", "handler"],
+)
+
+IDUNN_WIKI_EXCEPTIONS_COUNT = Counter(
+    "idunn_exceptions_count",
```
antoine-de: aren't them all about wikipedia ? if so it should be in the name, else you should change the name of the variable

idunn/blocks/wikipedia.py

idunn/utils/prometheus.py
jbgriesner: done

idunn/utils/prometheus.py
jbgriesner: done

antoine-de:
you should add an unhandled error
https://github.com/QwantResearch/idunn/blob/master/idunn/utils/logging.py#L34

PR #31, opened 2018-09-21
Grafana dashboard for prometheus metrics

antoine-de:
Add a dev prometheus/grafana setup to see the metrics

These are dev prometheus setup to see Idunn's metric.
This is absolutely not to be used in production!

To test it, you need to run the idunn's docker-compose with the additional prometheus compose. 
In the main idunn's directory:

`docker-compose -f docker-compose.yml -f prometheus/prometheus-compose.yml up --build`

Then you can query idunn on `localhost:5000` and access graphana on `http://localhost:3000/d/idunn_1/idunn-dashboard`.

(graphana's accesses are user: `admin`, password: `admin`).

If you change the dashboard, don't forget to export to json  and change the file `graphana_dashboards.json`.

NOTE: the dashboard is not complete, lots of metrics will need to be added after #30 

the dashboard look like this:
image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     build: .
     command: "--workers=4"
     environment:
-     - IDUNN_ES_URL=
+     - IDUNN_MIMIR_ES=
      - IDUNN_WIKI_API_REDIS_URL=idunn-redis:6379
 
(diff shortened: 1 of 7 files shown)


antoine-de:
I updated the dashboard after #30

PR #34, opened 2018-09-28
Catch Redis interruption

jbgriesner:
RedisError was catch in the rate limiter and in the init_cache() but not in the cache.
This PR just proposes to catch it also in the cache in order to prevent any Redis interruption.

### DIFF
--- a/idunn/blocks/wikipedia.py
+++ b/idunn/blocks/wikipedia.py
@@ -140,23 +140,22 @@ class WikipediaCache:
     _connection = None
 
     @classmethod
-    def set_value(cls, f, key, json_result):
+    def set_value(cls, key, json_result):
         try:
             cls._connection.set(key, json_result, ex=cls._expire)
             return True
         except RedisError:
-            logging.error("Got a RedisError in {}".format(f.__name__), exc_info=True)
+            logging.exception("Got a RedisError")
             return False
 
     @classmethod
-    def get_value(cls, key, f):
+    def get_value(cls, key):
         try:
             value_stored = cls._connection.get(key)
             return value_stored
         except RedisError:
-            logging.error("Got a RedisError in {}".format(f.__name__), exc_info=True)
+            logging.exception("Got a RedisError")
             return None
-
 
     @classmethod
     def init_cache(cls):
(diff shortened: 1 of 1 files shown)


idunn/blocks/wikipedia.py
```diff
@@ -180,14 +180,23 @@ def with_cache(*args, **kwargs):
...
             if cls._connection is not False:
-                value_stored = cls._connection.get(key)
-                if value_stored is not None:
-                    result = json.loads(value_stored.decode('utf-8'))
+                try:
+                    value_stored = cls._connection.get(key)
+                except RedisError:
+                    logging.error("Got a RedisError in {}".format(f.__name__), exc_info=True)
+                    return None
                 else:
```
antoine-de: this else is useless

idunn/blocks/wikipedia.py
```diff
@@ -180,14 +180,23 @@ def with_cache(*args, **kwargs):
             otherwise we bypass it
             """
             if cls._connection is not False:
-                value_stored = cls._connection.get(key)
-                if value_stored is not None:
-                    result = json.loads(value_stored.decode('utf-8'))
+                try:
+                    value_stored = cls._connection.get(key)
```
antoine-de: i'd prefer a `get` / `set` method in the class to hide this

idunn/blocks/wikipedia.py
```diff
@@ -180,14 +180,23 @@ def with_cache(*args, **kwargs):
...
+                        result = json.loads(value_stored.decode('utf-8'))
+                    else:
+                        result = f(*args, **kwargs)
+                        json_result = json.dumps(result)
+                        try:
+                            cls._connection.set(key, json_result, ex=cls._expire)
+                        except RedisError:
+                            logging.error("Got a RedisError in {}".format(f.__name__), exc_info=True)
+                            return None
+                    return result
```
antoine-de: in general i'd prefer early return in this function

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py

idunn/blocks/wikipedia.py
```diff
@@ -199,14 +198,13 @@ def with_cache(*args, **kwargs):
...
-                else:
-                    result = f(*args, **kwargs)
-                    json_result = json.dumps(result)
-                    if not cls.set_value(f, key, json_result):
-                        return None
+                    return json.loads(value_stored.decode('utf-8'))
+                result = f(*args, **kwargs)
+                json_result = json.dumps(result)
+                if not cls.set_value(key, json_result):
+                    return None
```
antoine-de: if we can't write the response in the cache, won't we cant to return this response anyway ?

jbgriesner: indeed you're right,
done

PR #35, opened 2018-10-16
New endpoint: /places/

jbgriesner:
This PR adds the endpoint "/v1/places/<id>" to returns:
- admin
- street
- address
- poi

The endpoint "/v1/pois/" is deprecated by this PR.

### DIFF
--- a/idunn/api/places.py
+++ b/idunn/api/places.py
@@ -1,21 +1,26 @@
-from apistar.exceptions import NotFound
+import logging
+from apistar.exceptions import NotFound, BadRequest
 from elasticsearch import Elasticsearch
 
 from idunn.api.poi import POI
 from idunn.api.pois import fetch_es_poi
 from idunn.api.place import Place, Admin, Street, Address
 from idunn.utils.settings import Settings
+from idunn.utils.index_names import IndexNames
 
-def get_index(type, settings):
-    return {
-        "admin": settings['PLACE_ADMIN_INDEX'],
-        "street": settings['PLACE_STREET_INDEX'],
-        "address": settings['PLACE_ADDRESS_INDEX'],
-        "poi": settings['PLACE_POI_INDEX'],
-    }.get(type, "_all")
+logger = logging.getLogger(__name__)
 
+def fetch_es_place(id, es, indices, type) -> list:
+    if type is None:
+        index_name = "munin"
+    elif type not in indices:
+        raise BadRequest(
+            status_code=400,
+            detail={"message": f"Wrong type parameter: type={type}"}
+        )
+    else:
+        index_name = indices.get(type)
 
-def fetch_es_place(id, es, settings, index_name) -> dict:
     es_places = es.search(index=index_name,
         body={
             "filter": {
@@ -25,42 +30,27 @@ def fetch_es_place(id, es, settings, index_name) -> dict:
 
     es_place = es_places.get('hits', {}).get('hits', [])
     if len(es_place) == 0:
-        raise NotFound(detail={'message': f"place '{id}' not found"})
+        raise NotFound(detail={'message': f"place '{id}' not found with type='{type}'"})
 
     return es_place
 
-def load_admin(es_place, lang, settings):
-    return Admin.load_place(es_place, lang, settings)
-
-def load_street(es_place, lang, settings):
-    return Street.load_place(es_place, lang, settings)
-
-def load_address(es_place, lang, settings):
-    return Address.load_place(es_place, lang, settings)
-
-def load_poi(es_place, lang, settings):
-    properties = {p.get('key'): p.get('value') for p in es_place.get('properties')}
-    es_place['properties'] = properties
-    return POI.load_place(es_place, lang, settings)
-
-def get_place(id, es: Elasticsearch, settings: Settings, lang=None, type=None) -> Place:
+def get_place(id, es: Elasticsearch, indices: IndexNames, settings: Settings, lang=None, type=None) -> Place:
     if not lang:
         lang = settings['DEFAULT_LANGUAGE']
     lang = lang.lower()
 
-    index_name = get_index(type, settings)
-    es_place = fetch_es_place(id, es, settings, index_name)
+    es_place = fetch_es_place(id, es, indices, type)
 
-    loader = {
-        "munin_adm": load_admin,
-        "munin_str": load_street,
-        "munin_add": load_address,
-        "munin_poi": load_poi,
+    places = {
+        "admin": Admin,
+        "street": Street,
+        "addr": Address,
+        "poi": POI,
     }
 
-    place = loader.get(es_place[0].get('_index')[:9])
+    loader = places.get(es_place[0].get('_type'))
 
-    if place is None:
+    if loader is None:
         return None
 
-    return place(es_place[0]['_source'], lang, settings)
+    return loader.load_place(es_place[0]['_source'], lang, settings)
--- a/idunn/api/places.py
+++ b/idunn/api/places.py
@@ -31,7 +31,7 @@ def fetch_es_place(id, es, indices, type) -> list:
 
     es_place = es_places.get('hits', {}).get('hits', [])
     if len(es_place) == 0:
-        raise NotFound(detail={'message': f"place '{id}' not found with type='{type}'"})
+        raise NotFound(detail={'message': f"place {id} not found with type={type}"})
 
     return es_place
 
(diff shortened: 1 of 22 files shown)


tests/test_places.py
```diff
@@ -0,0 +1,121 @@
...
+    assert response.status_code == 200
+    assert response.headers.get('Access-Control-Allow-Origin') == '*'
+
+    resp = response.json()
+
+    assert resp["id"] == "35460343"
+    assert resp["name"] == "9a Birnenweg"
+    assert resp["label"] == "9a Birnenweg"
+
+def ko_test_basic_query_address(address_du_moulin):
```
jbgriesner: This test fails: it returns a 404.
However the query to idunn for this address:
`curl localhost:5000/v1/places/"addr:5.108632;48.810273"?lang=fr`
returns the correct response: 
`{
  "type": "address",
  "id": "addr:5.108632;48.810273",
  "name": "4 Rue du Moulin",
  "label": "4 Rue du Moulin (Val-d'Ornain)"
}
`
I tried to investigate but everything looks ok, so Idk...

idunn/api/places.py
```diff
@@ -0,0 +1,66 @@
...
+def get_place(id, es: Elasticsearch, settings: Settings, lang=None, type=None) -> Place:
+    if not lang:
+        lang = settings['DEFAULT_LANGUAGE']
+    lang = lang.lower()
+
+    index_name = get_index(type, settings)
+    es_place = fetch_es_place(id, es, settings, index_name)
+
+    loader = {
+        "munin_adm": load_admin,
```
jbgriesner: Here I should have used the settings indices names instead of raw strings

idunn/api/places.py
```diff
@@ -0,0 +1,66 @@
...
+from idunn.api.place import Place, Admin, Street, Address
+from idunn.utils.settings import Settings
+
+def get_index(type, settings):
+    return {
+        "admin": settings['PLACE_ADMIN_INDEX'],
+        "street": settings['PLACE_STREET_INDEX'],
+        "address": settings['PLACE_ADDRESS_INDEX'],
+        "poi": settings['PLACE_POI_INDEX'],
+    }.get(type, "_all")
```
antoine-de: you can put this dict as a global variable.

And:
* we should never you `_all` but `munin` if we want all the indexes. else if the ES cluster is used for other thing we'll get strange results.
* I'm not a fan of getting all the indexes if the type is not known. if the index is `bob` I want an error, (I think it's on `type=None` that we want all indexes).

idunn/api/places.py
```diff
@@ -0,0 +1,66 @@
...
+    return Admin.load_place(es_place, lang, settings)
+
+def load_street(es_place, lang, settings):
+    return Street.load_place(es_place, lang, settings)
+
+def load_address(es_place, lang, settings):
+    return Address.load_place(es_place, lang, settings)
+
+def load_poi(es_place, lang, settings):
+    properties = {p.get('key'): p.get('value') for p in es_place.get('properties')}
```
antoine-de: do we want this only for poi ?

idunn/api/places.py
```diff
@@ -0,0 +1,66 @@
...
+    es_place = fetch_es_place(id, es, settings, index_name)
+
+    loader = {
+        "munin_adm": load_admin,
+        "munin_str": load_street,
+        "munin_add": load_address,
+        "munin_poi": load_poi,
+    }
+
+    place = loader.get(es_place[0].get('_index')[:9])
```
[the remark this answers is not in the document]
antoine-de: why don't you use the type of the objet ?

tests/test_places.py
```diff
@@ -0,0 +1,121 @@
...
+from app import app
+import pytest
+import json
+import os
+
+"""
+    This module tests basic query against the endpoint '/places/'
+"""
+
+def load_place(file_name, mimir_client, index_name, _doc_type):
```
antoine-de: why *_*doc_type ?

antoine-de: acn't you guess the index_name by the doc_type ?

tests/test_places.py
```diff
@@ -0,0 +1,121 @@
...
+
+@pytest.fixture(scope="session")
+def orsay_museum(mimir_client, init_indices):
+    """
+    fill elasticsearch with the orsay museum
+    """
+    return load_place('orsay_museum.json', mimir_client, 'munin_poi', 'poi')
+
+
+def test_basic_query_admin(admin_goujounac):
```
antoine-de: It's seems to work but I'm not a fan of this mechanism, you create side effects by loading stuff in ES and not cleaning afterward.
if you run this test first then `test_basic_query_street`, in the second test, ES will have `admin_goujounac` and `street_birnenweg`.

either always add all 3 objects of clean afterward*.

Side effects in tests can create very hard to debug king of bugs

tests/test_places.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+    assert response.status_code == 200
+    assert response.headers.get('Access-Control-Allow-Origin') == '*'
+
+    resp = response.json()
+
+    assert resp["id"] == "35460343"
+    assert resp["name"] == "9a Birnenweg"
+    assert resp["label"] == "9a Birnenweg"
```
antoine-de: It would be better to have a label different than the name, this way we are sure we did not screwd up between the 2

tests/test_places.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    fill elasticsearch with the orsay museum
+    """
+    return load_place('orsay_museum.json', mimir_client, 'munin_poi', 'poi')
+
+
+def test_basic_query_admin(admin_goujounac):
+    client = TestClient(app)
+    response = client.get(
+        url=f'http://localhost/v1/places/{admin_goujounac}?lang=fr',
```
antoine-de: you newer test with a `type` param, all tests should be done twice

and you should test with a `type=bob`

jbgriesner: done

tests/test_places.py
jbgriesner: done

tests/test_places.py
jbgriesner: done

tests/test_places.py
jbgriesner: yup, done

jbgriesner: removed

idunn/api/places.py
jbgriesner: yes

idunn/api/places.py
jbgriesner: ok, done

idunn/api/places.py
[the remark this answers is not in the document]
jbgriesner: yes done ^^

idunn/api/places.py
```diff
@@ -0,0 +1,66 @@
...
+def get_index(type, settings):
+    return {
+        "admin": settings['PLACE_ADMIN_INDEX'],
+        "street": settings['PLACE_STREET_INDEX'],
+        "address": settings['PLACE_ADDRESS_INDEX'],
+        "poi": settings['PLACE_POI_INDEX'],
+    }.get(type, "_all")
+
+
+def fetch_es_place(id, es, settings, index_name) -> dict:
```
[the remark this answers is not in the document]
jbgriesner: :+1:

idunn/api/poi.py
```diff
@@ -84,6 +87,15 @@ class POI(types.Type):
...
     blocks = BlocksValidator(allowed_blocks=BLOCKS_ORDER)
 
+    @classmethod
+    def load_place(cls, es_place, lang, settings):
+        properties = {p.get('key'): p.get('value') for p in es_place.get('properties')}
+        es_place['properties'] = properties
+        return cls.load_poi(es_place, lang)
+
+    """
+    @deprecated method
```
antoine-de: why is it deprecated ?

idunn/api/places.py
```diff
@@ -49,6 +52,8 @@ def get_place(id, es: Elasticsearch, indices: IndexNames, settings: Settings, la
     loader = places.get(es_place[0].get('_type'))
 
     if loader is None:
+        prometheus.exception("FoundPlaceWithWrongType {}".format(es_place[0].get('_type')))
```
antoine-de: I don't think we want the `type` in the prometheus exception, we want an aggregated metric

idunn/api/places.py
```diff
@@ -49,6 +52,8 @@ def get_place(id, es: Elasticsearch, indices: IndexNames, settings: Settings, la
...
+        logger.error("The place with the id {} has a wrong type".format(id), exc_info=True)
```
antoine-de: we don't want the `exc_info` there

antoine-de: and we want the type here :wink:

PR #37, opened 2018-10-24
limit wiki size

jbgriesner:
Just limit the size of the wiki description to 325 chars.
No test since the changes are minimum.


idunn/blocks/wikipedia.py
```diff
@@ -402,8 +404,10 @@ def from_es(cls, es_poi, lang):
             key = GET_SUMMARY + "_" + wiki_title + "_" + lang
             wiki_summary = WikipediaCache.cache_it(key, cls._wiki_session.get_summary)(wiki_title, lang=lang)
             if wiki_summary:
+                extract = wiki_summary.get("extract", "")
```
antoine-de: a function would be nice

PR #40, opened 2018-10-25
Upgrade hoh 1.0

jbgriesner:
Just upgrade the library used to parse opening hours and remove old patchs.


amatissart:
Thanks for adding a test :+1: 

A new version [osm-humanized-opening-hours==1.0.0b2](https://github.com/rezemika/humanized_opening_hours/releases/tag/v1.0.0b2) of the library has just been released, and includes all recent fixes. We should use it instead of depending on a specific commit, and that will be good to merge !

Pipfile
```diff
@@ -17,13 +17,12 @@ requests = "==2.*"
...
 babel = "*"
 pybreaker = "*"
-osm-humanized-opening-hours = "==0.6.2"
 redis = "*"
 python-redis-rate-limit = "*"
-lark-parser = "<0.6"
 prometheus-client = "*"
 apistar-prometheus = "*"
 python-json-logger = "*"
+osm-humanized-opening-hours = {editable = true, ref = "2fde45bec", git = "https://github.com/rezemika/humanized_opening_hours.git"}
```
amatissart: We can directly use `osm-humanized-opening-hours==1.0.0b2` now that it has been released

PR #43, opened 2018-11-30
Solar hours + 24/7

jbgriesner:
Just makes the 2 opening-hours of these 2 places correct in idunn:
- `"sunrise-sunset"` : https://www.openstreetmap.org/way/5095262
- `"24/7"` : https://www.openstreetmap.org/node/2055892768

### DIFF
--- a/idunn/blocks/opening_hour.py
+++ b/idunn/blocks/opening_hour.py
@@ -104,11 +104,9 @@ class OpeningHourBlock(BaseBlock):
         # This is why we replace the timezone info until this problem is fixed in the library.
         try:
             nt = oh.next_change(dt=poi_dt.replace(tzinfo=None))
-        except NextChangeRecursionError:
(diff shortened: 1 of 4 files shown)


idunn/blocks/opening_hour.py
```diff
@@ -66,30 +61,55 @@ def from_es(cls, es_poi, lang):
...
+                days=cls.get_days(oh, dt=poi_dt)
+            )
+
         # The current version of the hoh lib doesn't allow to use the next_change() function
         # with an offset aware datetime.
         # This is why we replace the timezone info until this problem is fixed in the library.
         try:
             nt = oh.next_change(dt=poi_dt.replace(tzinfo=None))
+        except NextChangeRecursionError:
+            logger.info("NextChangeRecursionError: Failed to compute next transition for poi %s", es_poi.get('id'), exc_info=True)
```
amatissart: Is not that exception already handled by `HOHError` ?  
Actually the root cause of the 500 error is not fixed: we need to return when the exception is catched (otherwise `nt` is not defined)

PR #44, opened 2018-12-12
More infos panel

jbgriesner:
Basic refactor + more infos to admin/street/address/poi panels.
work in progress...


idunn/places/address.py
```diff
@@ -0,0 +1,22 @@
...
+        raw_address = es_place.get('address') or {}
+        address = build_address(raw_address)
+
+        return cls(
+            id=es_place.get('id', ''),
+            name=es_place.get('name', ''),
+            local_name=es_place.get('name'),
+            class_name=es_place.get('poi_class'),
+            subclass_name=es_place.get('poi_subclass'),
+            geometry=get_geom(es_place, 'Point'),
```
amatissart: we shouldn't hard-code `Point`. And I think we could always return the whole geometry object for any type. (Even if a center is not computed for non-point geometry for now)

tests/test_full.py
```diff
@@ -41,10 +41,25 @@ def test_full():
...
+            ]
         },
+        "label": "62B Rue de Lille (Paris)",
         "address": {
-            "label": "62B Rue de Lille (Paris)"
+            "id": "addr:2.326285;48.859635",
+            "name": "62B Rue de Lille",
+            "house_number": None,
+            "label": "62B Rue de Lille (Paris)",
+            "zip_codes": None,
```
amatissart: to replace with a `postcode` and its value

tests/test_places.py
```diff
@@ -67,7 +68,6 @@ def test_basic_query_street():
 
     assert resp["id"] == "35460343"
     assert resp["name"] == "9a Birnenweg"
-    assert resp["label"] == "9a Birnenweg (Label)"
```
amatissart: These tests should specify the whole response, to prevent regression in the future.

tests/fixtures/basket_ball.json
```diff
@@ -7,7 +7,20 @@
     "label": "62B Rue de Lille (Paris)",
     "name": "62B Rue de Lille",
     "postcode": "75007",
-    "street": "Rue de Lille",
```
amatissart: We should check these fixtures. I don't think that `postcode` should appear in the documents that are inserted into ES.

idunn/places/poi.py
```diff
@@ -0,0 +1,68 @@
+from .place import Place
+from idunn.api.utils import build_blocks, get_geom, get_name
+
+
+class POI(Place):
+    PLACE_TYPE = 'poi'
+
+    @classmethod
+    def build_address(cls, es_place):
```
amatissart: I don't understand, why a different implementation is needed, compared to `Place.build_address`

idunn/places/place.py
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                    admins.append(admin)
+            postcode = raw_address.get("zip_codes") or es_place.get("zip_codes")
+            if postcode is not None:
+                if isinstance(postcode,list):
+                    postcode = ';'.join(postcode)
+            return {
+                "id": raw_address.get("id"),
+                "name": raw_address.get("name"),
+                "house_number": raw_address.get("house_number"),
```
amatissart: bragi uses `housenumber` (without underscore) in the response.  
It would be helpful to do the same here.

tests/test_full.py
```diff
@@ -41,10 +41,26 @@ def test_full():
...
         },
+        "label": "62B Rue de Lille (Paris)",
         "address": {
-            "label": "62B Rue de Lille (Paris)"
+            "admins": [],
+            "id": "addr:2.326285;48.859635",
+            "name": "62B Rue de Lille",
+            "house_number": "62B",
+            "label": "62B Rue de Lille (Paris)",
+            "postcode": "75007;75008",
```
amatissart: We would expect a single postcode in this case, with a specific address with house number.  
When reading this test, it's hard to say if this is related to the implementation or due to an unrealistic fixture.

toto.json
```diff
@@ -0,0 +1 @@
+
```
amatissart: this file is hopefully not required :smile:

tests/test_places.py
amatissart: Still needed for request about admins and addresses

tests/test_places.py
```diff
@@ -75,64 +114,35 @@ def test_basic_query_street():
...
-                {
-                    "id": "admin:osm:relation:367854",
-                    "label": "Sankt Lorenz S\u00fcd, L\u00fcbeck, Schleswig-Holstein, Deutschland",
-                    "name": "Sankt Lorenz S\u00fcd",
-                    "level": 9,
-                    "postcode": []
-                    }
-                ]
-            },
+                { "id": "admin:osm:relation:27027", "label": "L\u00fcbeck, Schleswig-Holstein, Deutschland", "name": "L\u00fcbeck", "level": 6, "postcode": [] },
```
amatissart: `postcode` :arrow_right: `postcodes`  
`level` to replace with `class_name`

tests/test_places.py
```diff
@@ -145,11 +155,44 @@ def test_basic_query_address():
...
+        'subclass_name': 'address',
+        'geometry': {
+            'type': 'Point',
+            'coordinates': [5.108632, 48.810273],
+            'center': [5.108632, 48.810273]
+        },
+        'address': {
+            'id': None,
+            'name': None,
+            'housenumber': None,
```
amatissart: these fields should not be missing

tests/test_places.py
```diff
@@ -160,15 +203,47 @@ def test_basic_query_poi():
...
+        'subclass_name': 'museum',
+        'geometry': {
+            'type': 'Point',
+            'coordinates': [2.3250037768187326, 48.86618482685007],
+            'center': [2.3250037768187326, 48.86618482685007]
+        },
+        'address': {
+            'id': 'addr:2.326285;48.859635',
+            'name': '62B Rue de Lille',
+            'housenumber': None,
```
amatissart: this field should be `"62B"`

PR #45, opened 2018-12-14
Add images

jbgriesner:
return thumbrized image urls


idunn/blocks/images.py
```diff
@@ -0,0 +1,131 @@
...
+
+        size = int(width) * int(height)
+
+        hashURLpart = f"{hash[0]}/{hash[1]}/{hash[2:]}"
+
+        filename = source.split('/')[-1]
+        if not bool(re.match("^.*\.(jpg|jpeg|png|gif)$", filename)):
+            filename += ".jpg"
+
+        params = f"u={urllib.parse.quote(source)}&q={1 if displayErrorImage else 0}&b={1 if bestFit else 0}&p={1 if progressive else 0}&a={1 if animated else 0}"
```
amatissart: Again, that needs to be tested. But according to https://docs.python.org/3/library/urllib.parse.html#url-quoting `quote` does not encode `/` characters. Is that a problem ? 
Using [`urllib.parse.urlencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) (that accepts a dict as input) may be more suitable

idunn/blocks/images.py
```diff
@@ -0,0 +1,131 @@
...
+                    resp = wiki_es.search(
+                        index=wiki_index,
+                        body={
+                            "filter": {
+                                "term": {
+                                    "wikibase_item": wikidata_id
+                                }
+                            }
+                        }
+                    ).get('hits', {}).get('hits', [])
```
amatissart: We should try not to query the ES a second time, after the wiki block.  
And this implementation should not be duplicated anyway.  
Why not persist the wiki_es response in a wrapper class for es_poi ? Then it could be used by both blocks with a single query.

idunn/blocks/images.py
```diff
@@ -0,0 +1,131 @@
...
+
+                images = None
+                if not ThumbrHelper.use_thumbr():
+                    image_wiki = Image(url=urllib.parse.quote(raw_url), alt=place_name, credits="wikimedia")
+                    images = [image_wiki]
+                else:
+                    thumbr_url_large = ThumbrHelper.get_url_remote_thumbnail(
+                        source=raw_url,
+                        width=30,
+                        height=30
```
amatissart: we need to think about how to set these values.  
I suppose the ratio of images is not fixed. Is the size of the image present in the response ?  
Should we force a ratio on our side, or set these values dynamically ?  
To discuss with the frontend ; it's also possible to let the frontend set the `bestFit` value to truncate the image if needed.

idunn/utils/default_settings.yaml
```diff
@@ -43,3 +43,7 @@ PLACE_ADDRESS_INDEX: "munin_addr"
 PLACE_POI_INDEX: "munin_poi"
 
 WIKI_DESC_MAX_SIZE: 325 # max size allowed to the description of the wiki block
+
+THUMBR_SALT:
+USE_THUMBR: True # Set to False if you want to keep raw image urls instead of Thumbr
+THUMBR_DOMAINS: "s1.qwant.plive,s2.qwant.plive"
```
amatissart: By default `qwant.com` should be used instead

idunn/blocks/images.py
```diff
@@ -0,0 +1,131 @@
...
+        salt = cls.get_salt()
+        token = f"{source}{width}x{height}{salt}"
+        hash = hashlib.sha256(bytes(token, encoding='utf8')).hexdigest()
+        domain = cls.get_domain(hash)
+
+        size = int(width) * int(height)
+
+        hashURLpart = f"{hash[0]}/{hash[1]}/{hash[2:]}"
+
+        filename = source.split('/')[-1]
```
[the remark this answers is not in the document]
jbgriesner: Actually these url cannot contain params :)

idunn/blocks/wikipedia.py
```diff
@@ -330,6 +336,8 @@ def get_wiki_info(cls, wikidata_id, lang, wiki_index):
 
         wiki = resp[0]['_source']
 
+        es_poi.wiki_resp = wiki
```
amatissart: I am not fond of this solution, especially because it requires this block to be instantiated before other blocks that depends on wikidata details.   
Would it be possible to call `WikidataConnector` directly from `PlaceData` ?

idunn/utils/default_settings.yaml
amatissart: Actually I meant `s1.qwant.com,s2.qwant.com` ^^

PR #48, opened 2019-02-06
Implement /places with bbox and class,subclass filters

jbgriesner:
Add a route to query all POI that match:
- a given list of categories
- a given bounding box

We do not return more than a max number of POI, ranked by ES weight.


idunn/api/places_list.py
```diff
@@ -0,0 +1,69 @@
...
+        send_400(f"the bounding box cannot be None")
+
+    bbox = [float(c) for c in bbox.split(",")]
+    left, bot, right, top = bbox[0], bbox[1], bbox[2], bbox[3]
+
+    if not all(bbox):
+        send_400(f"the bounding box is not complete")
+    if left > right or bot > top:
+        send_400(f"the bounding box is not valid")
+    if left - right > 0.181:
```
jbgriesner: this should be defined somewhere else I guess

idunn/api/places_list.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+        bbox = left,bottom,right,top
+    """
+    places_list = []
+    params = dict(query_params)
+
+    bbox = bbox_or_400(verbosity, params.get('bbox'))
+
+    try:
+        categories = params['categories']
```
jbgriesner: I didn't find a way to make apistar fetch the categories list directly, so I'll parse it manually as a string, which is ugly

idunn/api/utils.py
```diff
@@ -42,6 +50,68 @@ def fetch_es_poi(id, es) -> dict:
     result['properties'] = properties
     return result
 
+def fetch_bbox_places(bbox, es, indices, categories, max_size) -> list:
+    left, bot, right, top = bbox[0], bbox[1], bbox[2], bbox[3]
+
+    categories = re.findall(r'\((\w*?,\w*?)\)', categories)
+    classes, subclasses, class_sub = [], [], []
```
jbgriesner: `classes` contains the cases of the categories of the form `(<class>,_any)`
`subclasses` contains the cases of the categories of the form `(_any,<subclass>)`
`class_sub` contains the categories without any occurrence of `_any`

idunn/api/utils.py
```diff
@@ -42,6 +50,68 @@ def fetch_es_poi(id, es) -> dict:
...
+            classes.append(pair.split(",")[0])
+        elif pair.split(",")[0] == "_any":
+            subclasses.append("subclass_" + pair.split(",")[1])
+            subclasses.append(pair.split(",")[1])
+        else:
+            class_sub.append("class" + pair.split(",")[0] + " subclass_" + pair.split(",")[1])
+            class_sub.append(pair.split(",")[0] + " " + pair.split(",")[1])
+
+    should_terms = []
+    for term in [classes, subclasses, class_sub]:
```
jbgriesner: this loop purpose is to avoid any empty list in the queries below, however I have no doubt there are ways with ES to avoid this

tests/test_categories.py
```diff
@@ -0,0 +1,161 @@
...
+        mimir_client.index(index='munin_poi',
+                        body=poi,
+                        doc_type='poi',
+                        id=poi_id,
+                        refresh=True)
+
+@pytest.fixture(autouse=True)
+def load_all(mimir_client, init_indices):
+    """
+    fill elasticsearch with 4 POI:
```
jbgriesner: these fixtures should be updated to take into account of the fafnir last changes https://github.com/QwantResearch/fafnir/pull/25

idunn/api/places_list.py
```diff
@@ -0,0 +1,69 @@
...
+from idunn.places import Place, Admin, Street, Address, POI
+from idunn.api.utils import get_geom, get_name, fetch_bbox_places, LONG, SHORT, DEFAULT_VERBOSITY_LIST, send_400
+
+from apistar import http
+
+logger = logging.getLogger(__name__)
+
+VERBOSITY_LEVELS = [LONG, SHORT]
+
+def bbox_or_400(verbosity, bbox):
```
amatissart: this function should be renamed.  
Or even better, why not validating each parameter separately ?

idunn/api/places_list.py
amatissart: why this value specifically ?

idunn/api/places_list.py
```diff
@@ -0,0 +1,69 @@
...
+    if verbosity not in VERBOSITY_LEVELS:
+        send_400(f"verbosity {verbosity} does not belong to the set of possible verbosity values={VERBOSITY_LEVELS}")
+    if not bbox:
+        send_400(f"the bounding box cannot be None")
+
+    bbox = [float(c) for c in bbox.split(",")]
+    left, bot, right, top = bbox[0], bbox[1], bbox[2], bbox[3]
```
amatissart: the length of the list should be validated too. What if there are less or more than 4 values ?

idunn/api/utils.py
```diff
@@ -18,6 +20,14 @@
     ]
 }
 
+ANY = "_any"
+
+def send_400(msg):
+    raise BadRequest(
```
amatissart: it would be more explicit to define that as a new exception class, derived from `BadRequest`.

idunn/api/utils.py
amatissart: I think a single list would work here

idunn/api/utils.py
```diff
@@ -42,6 +52,68 @@ def fetch_es_poi(id, es) -> dict:
...
+            class_sub.append("class" + pair.split(",")[0] + " subclass_" + pair.split(",")[1])
+            class_sub.append(pair.split(",")[0] + " " + pair.split(",")[1])
+
+    should_terms = []
+    for term in [classes, subclasses, class_sub]:
+        if term != []:
+            should_terms.append(
+                {
+                    "terms": {
+                        "poi_type.name": term
```
amatissart: i am not sure `terms` works here, especially with multiple words in one of `class_sub` values

tests/test_full.py
```diff
@@ -30,152 +19,43 @@ def test_full():
     resp = response.json()
 
     assert resp == {
-        "type": "poi",
```
amatissart: this test should not change. (As discussed, due to duplicate id in fixtures)

PR #51, opened 2019-02-28
Add the 'category' filter

jbgriesner:
After the `raw_filter` param this PR adds the `category` param with few validation rules:
- exactly one (not more not less than one) of these two params must be provided 
- the provided categories must be in the categories listed in the `categories.yml` file


idunn/utils/categories.py
```diff
@@ -0,0 +1,29 @@
...
+from .settings import Settings
+from elasticsearch import Elasticsearch
+from .settings import _load_yaml_file
+
+class Categories(dict):
+    """
+        Class to handle the possible categories of places to fetch
+    """
+
+class CategoriesComponent(Component):
```
jbgriesner: This class is very similar to `IndexNamesSettingsComponent` and `SettingsComponent`.
What about using any abstraction to reduce code duplications ?

idunn/utils/categories.yml
```diff
@@ -0,0 +1,26 @@
+categories:
```
jbgriesner: Probably better formats are possible

idunn/api/places_list.py
```diff
@@ -4,28 +4,44 @@
...
 
 logger = logging.getLogger(__name__)
 
 VERBOSITY_LEVELS = [LONG, SHORT]
 
-MAX_WIDTH = 1.0 # max bbox longitude in degrees
+MAX_WIDTH = 1.0 # max bbox longitude in degrees
 MAX_HEIGHT = 1.0  # max bbox latitude in degrees
 
+ALL_CATEGORIES = None
```
jbgriesner: The purpose of this global variable is to access it into the validators easily

idunn/api/places_list.py
```diff
@@ -58,16 +74,39 @@ def max_size(cls, v):
...
+    def valid_category(cls, v):
+        if v not in ALL_CATEGORIES:
+            raise ValueError(f"Category \'{v}\' is invalid since it does not belong to the set of possible categories: {list(ALL_CATEGORIES.keys())}")
+        else:
+            v = ALL_CATEGORIES.get(v)
+        return v
+
+    @validator('category', whole=True)
+    def flat_categories(cls, v):
+        return [filt for filters in v for filt in filters]
```
jbgriesner: Since a single category is associated to *a list of raw_filters* (c.f. categories.yml) here the `category` is a list of validated lists : this is why we flat it

idunn/api/places_list.py
amatissart: instead of using a component, it would be more simple to read the categories from file here

idunn/utils/categories.yml
amatissart: to be more extendable: 
```
categories:
  restaurant:
    raw_filters:
       - "*,restaurant"
       - "*,cafe"
```

idunn/utils/default_settings.yaml
```diff
@@ -45,3 +45,5 @@ PLACE_POI_INDEX: "munin_poi"
 WIKI_DESC_MAX_SIZE: 325 # max size allowed to the description of the wiki block
 
 LIST_PLACES_MAX_SIZE: 50
+
+CATEGORIES_FILENAME: "categories.yml"
```
amatissart: i think we could drop this setting for now

tests/test_categories.py
```diff
@@ -286,3 +286,77 @@ def test_invalid_bbox():
             }
         ]
     }
+
+
+@freeze_time("2018-06-14 8:30:00", tz_offset=2)
+def test_category_and_raw_filter():
```
amatissart: one more test with a valid category, and it will be perfect

idunn/api/places_list.py
```diff
@@ -1,31 +1,52 @@
...
+class Categories:
+    _categories = None
+
+    @classmethod
+    def get_categories(cls):
+        if cls._categories is None:
+            categories_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../utils/categories.yml")
+            if categories_path:
+                categories = _load_yaml_file(categories_path)
+        return categories.get('categories')
```
amatissart: this is too complicated, and `_categories` is not used.
Why not removing this class, and simply use:
```python
def get_categories():
   categories_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../utils/categories.yml")
   return _load_yaml_file(categories_path)['categories']

all_categories = get_categories()
```
so that the yaml file is read only once.

idunn/api/urls.py
```diff
@@ -22,5 +23,6 @@ def get_api_urls(settings):
         Route('/status', 'GET', handler=get_status),
         Route('/pois/{id}', 'GET', handler=get_poi),
         Route('/places/{id}', 'GET', handler=get_place),
+        Route('/places/categories', 'GET', handler=get_all_categories),
```
amatissart: `/categories` would be more consistent, I think

idunn/utils/categories.yml
```diff
@@ -0,0 +1,30 @@
...
+    restaurant:
+        raw_filters:
+            - "*,restaurant"
+            - "*,cafe"
+            - "*,fast_food"
```
amatissart: Maybe we should rely on classes, instead of subclasses when possible. Both for retro-compatibility and maintainability.  
`restaurant,*`, `cafe,*`, etc

idunn/api/urls.py
amatissart: And `readme.md` is to update ;)

PR #57, opened 2019-04-25
Implement source param and a new verbosity when fetching multiple places

amatissart:
* `list` verbosity does not contain "information" block (including wikipedia info) contrary to `long`, and is used by default on GET /places?bbox=...
* `source` parameter can be used by the client to force the source selection. Eg: `?source=osm`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,11 +18,18 @@
 `GET /schema`
 
 The main endpoints are:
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -18,11 +18,18 @@
 `GET /schema`
 
 The main endpoints are:
-* `/v1/places/{place_id}?lang={lang}&type={type}&verbosity={verbosity}` to get the details of a place (admin, street, address or POI). The `type` parameter belongs to the set `{'admin', 'street', 'address', 'poi'}`. The `verbosity` parameter belongs to the set `{'long', 'short'}`. The default verbosity is `long`.
+* `/v1/places/{place_id}?lang={lang}&type={type}&verbosity={verbosity}` to get the details of a place
+(admin, street, address or POI).
+    * `type`: (optionnal) parameter belongs to the set `{'admin', 'street', 'address', 'poi'}`
```
jbgriesner: lets add the 'source' parameter here ?

PR #61, opened 2019-05-28
Handle phone numbers fully

GuillaumeGomez:


idunn/places/pj_poi.py
```diff
@@ -1,6 +1,8 @@
+import phonenumbers
```
jbgriesner: is this import really required ?

PR #65, opened 2019-06-11
Add diet and cuisine blocks

GuillaumeGomez:


idunn/blocks/services_and_information.py
```diff
@@ -94,11 +94,77 @@ def from_es(cls, es_poi, lang):
...
+    STATUS_ONLY = "only"
+
+    value = validators.String(
+        enum=[STATUS_YES, STATUS_NO, STATUS_ONLY]
+    )
+
+    @classmethod
+    def from_es(cls, info, lang):
+        info = info.split('=')
+        if info[0] not in ("pescetarian", "vegetarian", "lacto_vegetarian", "ovo_vegetarian",
```
GuillaumeGomez: To be reduced to "vegetarian", "vegan" and "gluten_free".

idunn/blocks/services_and_information.py
```diff
@@ -94,11 +94,77 @@ def from_es(cls, es_poi, lang):
...
+
+
+class DietBlock(BaseBlock):
+    BLOCK_TYPE = "diet"
+
+    diets = validators.Array(items=Diet)
+
+    @classmethod
+    def from_es(cls, es_poi, lang):
+        diet = es_poi.get("properties", {}).get("diet")
```
GuillaumeGomez: The key should be "diet:[diet_kind]".

idunn/blocks/services_and_information.py
```diff
@@ -94,11 +94,64 @@ def from_es(cls, es_poi, lang):
...
+
+
+class CuisineBlock(BaseBlock):
+    BLOCK_TYPE = "cuisine"
+    SUPPORTED_DIETS = ("vegetarian", "vegan", "gluten_free")
+
+    cuisines = validators.Array(items=Cuisine)
+    vegetarian_status = validators.Object()
+    vegan_status = validators.Object()
+    gluten_free_status = validators.Object()
```
amatissart: Any reason to use an object here ? 
Wouldn't `"vegetarian": STATUS_YES` be more simple, with a string value among all statuses ?

idunn/blocks/services_and_information.py
```diff
@@ -94,11 +94,64 @@ def from_es(cls, es_poi, lang):
...
+
+    @classmethod
+    def from_es(cls, es_poi, lang):
+        cuisine = es_poi.get("properties", {}).get("cuisine")
+
+        if cuisine is None:
+            return None
```
amatissart: I think we may want to return other details ("vegetarian", etc) that could be present, even if "cuisine" is absent.

tests/cuisine.py
```diff
@@ -0,0 +1,19 @@
+from app import app
+from idunn.blocks.services_and_information import CuisineBlock, Cuisine
+
+def test_internet_access_block():
```
amatissart: testcase to rename

GuillaumeGomez:
Updated.

PR #74, opened 2019-07-10
Validate URLs with float, fix bug in apistar injector

amatissart:
* Validate float values, and accept negative values using werkzeug syntax https://werkzeug.palletsprojects.com/en/0.15.x/routing/#werkzeug.routing.FloatConverter
* Work around bug in apistar that may cause the injector cache to be polluted with incomplete dependency resolution. Fixed using a custom `IdunnApp` class.

### DIFF
--- a/idunn/utils/app.py
+++ b/idunn/utils/app.py
@@ -1,0 +1,59 @@
+from apistar import App
+
+class IdunnApp(App):
+    def __call__(self, environ, start_response):
+        def get_default_state():
+            return {
+                'environ': environ,
+                'start_response': start_response,
+                'exc': None,
+                'app': self,
+                'path_params': None,
+                'route': None,
+                'response': None,
+            }
+
+        state = get_default_state()
+        method = environ['REQUEST_METHOD'].upper()
+        path = environ['PATH_INFO']
+
+        if self.event_hooks is None:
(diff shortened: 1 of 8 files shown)


idunn/utils/app.py
```diff
@@ -0,0 +1,59 @@
...
+            else:
+                funcs = (
+                    on_request +
+                    [route.handler, self.render_response] +
+                    on_response +
+                    [self.finalize_wsgi]
+                )
+            return self.injector.run(funcs, state)
+        except Exception as exc:
+            # This is the only change with the parent class.
```
GuillaumeGomez: Instead of keeping everything inside this exception block, since we return from there, why not remove one indent level and put `pass` and some logging into the exception block instead?

GuillaumeGomez:
Some code looking improvements to do but otherwise looks good. :)

idunn/utils/app.py
amatissart: I prefer to stick to the parent class implementation if you don't mind.
And `exc` is used in the block, so it looks more explicit this way.

GuillaumeGomez: I really don't like the fact that you have inner exception blocks in it. What about a function call instead?

amatissart: it should look cleaner now

PR #77, opened 2019-07-23
Add CORS support

GuillaumeGomez:
Linked to https://github.com/QwantResearch/erdapfel/pull/313/

### DIFF
--- a/idunn/api/places.py
+++ b/idunn/api/places.py
@@ -1,6 +1,7 @@
 import logging
 from elasticsearch import Elasticsearch
 from apistar.exceptions import BadRequest, NotFound
+from apistar.http import Response, Headers
 
 from idunn import settings
 from idunn.utils import prometheus
@@ -9,6 +10,7 @@ from idunn.places import Place, Admin, Street, Address, POI, Latlon
 from idunn.api.utils import fetch_es_place, DEFAULT_VERBOSITY, ALL_VERBOSITY_LEVELS
 from idunn.api.pages_jaunes import pj_source
 from .closest import get_closest_place
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


idunn/api/places.py
```diff
@@ -60,3 +64,14 @@ def get_place_latlon(lat: float, lon: float, es: Elasticsearch, lang=None, verbo
         closest_place = None
     place = Latlon(lat, lon, closest_address=closest_place)
     return place.load_place(lang, verbosity)
+
+
+def handle_option(id, es: Elasticsearch, indices: IndexNames, lang=None, type=None, verbosity=DEFAULT_VERBOSITY):
+    if settings.get('ENABLE_OPTION_REQUESTS', False) is True:
+        headers = {
+            'Access-Control-Allow-Origin': '*',
+            'Access-Control-Allow-Headers': '*',
```
amatissart: It's unclear if this wildcard value is well supported by browsers: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#Compatibility_notes

Maybe it would be safer to use the value set in `Access-Control-Request-Headers` request header

idunn/api/places.py
```diff
@@ -60,3 +64,14 @@ def get_place_latlon(lat: float, lon: float, es: Elasticsearch, lang=None, verbo
         closest_place = None
     place = Latlon(lat, lon, closest_address=closest_place)
     return place.load_place(lang, verbosity)
+
+
+def handle_option(id, es: Elasticsearch, indices: IndexNames, lang=None, type=None, verbosity=DEFAULT_VERBOSITY):
```
amatissart: most of these parameters can be removed

GuillaumeGomez:
Updated.

PR #80, opened 2019-08-09
Log custom headers

GuillaumeGomez:


idunn/api/places.py
```diff
@@ -29,7 +30,39 @@ def validate_lang(lang):
...
+                zoom = float(pos[2])
+                custom_data['lon'] = lon
+                custom_data['lat'] = lat
+                custom_data['zoom'] = zoom
+            except Exception as err:
+                logger.warning('Invalid data given through "X-QwantMaps-FocusPosition" header: ' + err)
+    if 'X-QwantMaps-Query' in headers
+        query = headers.get('X-QwantMaps-Query', '')
+        if len(query) > 0:
+            custom_data['query'] = query
```
amatissart: this header value is url-encoded and needs to be decoded

idunn/api/places.py
```diff
@@ -29,7 +30,39 @@ def validate_lang(lang):
...
+        if len(pos) == 3:
+            try:
+                lon = float(pos[0])
+                lat = float(pos[1])
+                zoom = float(pos[2])
+                custom_data['lon'] = lon
+                custom_data['lat'] = lat
+                custom_data['zoom'] = zoom
+            except Exception as err:
+                logger.warning('Invalid data given through "X-QwantMaps-FocusPosition" header: ' + err)
```
amatissart: To include details about the current exception, it's cleaner to pass `exc_info=True` instead of string concatenation

idunn/api/places.py
```diff
@@ -29,7 +30,39 @@ def validate_lang(lang):
...
+                ranking = int(ranking)
+                custom_data['ranking'] = ranking
+            except Exception as err:
+                logger.warning('Invalid data given through "X-QwantMaps-SuggestionRank" header: ' + err)
+    if 'X-QwantMaps-QueryLang' in headers
+        lang = headers.get('X-QwantMaps-QueryLang', '')
+        if len(lang) > 0:
+            custom_data['lang'] = lang
+    if len(custom_data) > 0:
+        logger.info('[get_place headers]: ' + json.dumps(custom_data))
```
amatissart: We should include the object in the structure of the json log entry, using `extra` keyword
```suggestion
        logger.info('Received details about user query' , extra={'user_selection':custom_data})
```

idunn/api/places.py
```diff
@@ -29,7 +30,39 @@ def validate_lang(lang):
     return lang.lower()
 
 
-def get_place(id, es: Elasticsearch, indices: IndexNames, lang=None, type=None, verbosity=DEFAULT_VERBOSITY) -> Place:
+def get_place(id, es: Elasticsearch, indices: IndexNames, headers: Headers, lang=None, type=None, verbosity=DEFAULT_VERBOSITY) -> Place:
+    custom_data = {}
+    if 'X-QwantMaps-FocusPosition' in headers:
```
amatissart: Parsing of the headers should be moved to a dedicated function

GuillaumeGomez:
Updated.

PR #81, opened 2019-08-14
Directions endpoint

amatissart:
Implements a new endpoint `/v1/directions/<start>;<end>`  
It is a wrapper for both APIs (walking / cycling / driving and public transport).

The response format is defined by a pydantic Model `DirectionsResponse`.  
More details are present in `doc/directions.md`


doc/directions.md
```diff
@@ -0,0 +1,2118 @@
...
+    }, 
+    "status": "success"
+}
+```
+
+### `publictransport`
+
+*Some geometries are simplified for readability*
+
+```json
```
GuillaumeGomez: You should put your code examples between `<details>` html tags. This is just too big as is.

idunn/directions/client.py
```diff
@@ -0,0 +1,102 @@
...
+                'language': lang,
+                'steps': 'true',
+                'alternatives': 'true',
+                'overview': 'full',
+                'geometries': 'geojson',
+                **extra
+            }
+        )
+
+        if 400 <= response.status_code < 500:
```
GuillaumeGomez: TIL

GuillaumeGomez: Isn't there a way to just check on the status code directly it's an error one?

idunn/directions/client.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        response.raise_for_status()
+
+        return DirectionsResponse(
+            status='success',
+            data=response.json()
+        ).dict()
+
+    def get_directions(self, from_loc, to_loc, mode, lang, **extra):
+        if mode in ('driving-traffic', 'driving', 'car'):
```
GuillaumeGomez: Instead of those if else, why not making a loop over a dict instead? It'd make future diffs simpler too.

idunn/directions/client.py
```diff
@@ -0,0 +1,102 @@
...
+        elif mode in ('publictransport', 'taxi', 'vtc', 'carpool'):
+            method = self.directions_combigo
+            mode = mode
+        else:
+            raise BadRequest(f'unknown mode {mode}')
+
+        return method(from_loc, to_loc, mode, lang, **extra)
+
+
+
```
GuillaumeGomez: That's a lot of empty lines.

GuillaumeGomez:
Needs confirmation through testing but seems globally good!

idunn/directions/client.py
amatissart: no, it does not seem that "requests" library provides any other method for that

doc/directions.md
amatissart: done

idunn/directions/client.py
amatissart: cleaned up

idunn/directions/models.py
```diff
@@ -0,0 +1,196 @@
...
+    tram = 'TRAM'
+    car_rental = 'CAR_RENTAL'
+    transfert = 'TRANSFERT'
+    subway = 'SUBWAY'
+    suburban_train = 'SUBURBAN_TRAIN'
+    seaplane = 'SEAPLANE'
+    helicopter = 'HELICOPTER'
+    funicular = 'FUNICULAR'
+    shuttle = 'SHUTTLE'
+    unknown = 'UNKNOW'
```
bbecquet: The final 'N' is missing, I guess that's a typo.

amatissart: That's actually how it is documented on https://maps2.combigo.com/v1/api-docs/#/ (model `TransportType`). Maybe a typo on their side though.

I've just added a comment.

PR #82, opened 2019-08-21
directions: Implement timeout and mapbox proxy

amatissart:
* Define request timeout on directions
* Implement mapbox proxy for development purposes

### DIFF
--- a/idunn/directions/client.py
+++ b/idunn/directions/client.py
@@ -1,22 +1,25 @@
 import requests
 import logging
-from idunn import settings
 from datetime import datetime, timedelta
-from .models import DirectionsResponse
 from apistar.http import JSONResponse
 from apistar.exceptions import BadRequest, HTTPException
+from idunn import settings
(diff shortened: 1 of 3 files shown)


idunn/directions/client.py
```diff
@@ -77,15 +126,20 @@ def directions_combigo(self, start, end, mode, lang, **kwargs):
...
 
+
     def get_directions(self, from_loc, to_loc, mode, lang, **extra):
+        default_method = self.directions_qwant
+        if self.MAPBOX_API_ENABLED:
+            default_method = self.directions_mapbox
+
         if mode in ('driving-traffic', 'driving', 'car'):
-            method = self.directions_qwant
+            method = default_method
```
GuillaumeGomez: Instead of repeating this line in all `if` blocks, why not setting it once before all `if else`?

amatissart: done

PR #84, opened 2019-08-22
combigo: Define supported languages

amatissart:


idunn/directions/client.py
```diff
@@ -9,6 +9,7 @@
 
 logger = logging.getLogger(__name__)
 
+COMBIGO_SUPPORTED_LANGUAGES = {'en','es','de','fr','it'}
```
GuillaumeGomez: You have the **_right_** to put whitespace between items you know. :)

amatissart: done :)

PR #86, opened 2019-09-12
add events

plavacquery:
mix events from several datasource,
add category (outing_type) for events


idunn/api/places_list.py
```diff
@@ -103,7 +112,15 @@ def valid_category(cls, v):
 
 
 class EventQueryParam(CommonQueryParam):
-    pass
+    outing_types: str = None
```
amatissart: why plural ? I think a single value is expected here.

Also, this parameter name is too specific in my opinion. On reflection, using `type` may be confusing, because `"type":"event"` is already present in the response. I would stick to `category`, similarly to the /places endpoint.

idunn/api/places_list.py
```diff
@@ -177,10 +194,20 @@ def get_events_bbox(bbox, query_params: http.QueryParams):
             detail={"message": e.errors()}
         )
 
+    print(params.outing_types)
+    current_outing_lang = params.outing_types
+
+    if params.outing_types:
+        lang = params.lang.split(',')[0]
+        current_outing_lang = params.outing_types.get(lang)
```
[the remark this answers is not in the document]
plavacquery: no it use for the translation in outing_types (category)

tests/test_environment_weather.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+
+        res = Weather.from_es(
+            Admin(testee),
+            lang='en'
+        )
+
+    assert res == Weather(**{
+        'temperature': 291.89,
+        'weatherPic': '01d',
```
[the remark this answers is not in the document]
plavacquery: It is the icon code value, do not find any better value

idunn/blocks/events.py
```diff
@@ -53,6 +53,7 @@ class DescriptionEvent(BaseBlock):
     description = validators.String(allow_null=True)
     free_text = validators.String(allow_null=True)
     price = validators.String(allow_null=True)
+    tags = validators.String(allow_null=True)
```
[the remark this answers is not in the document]
plavacquery: the tags are an entire string,redo it as a list

idunn/api/places_list.py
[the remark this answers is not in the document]
plavacquery: indeed it use for es query because, it find out the word concert, show, exhibition, .. in the appropriate language

idunn/api/places_list.py
plavacquery: ok I will redefine it as 'outing_category'

idunn/api/kuzzle.py
```diff
@@ -93,19 +93,38 @@ def fetch_event_places(self, bbox, size) -> list:
...
                             }
                         }
-                    },
-                    'must': {
-                        'range': {
-                            'date_end': {
-                                'gte': 'now/d',
-                                'lte': 'now+31d/d'
-                            }
-                        }
```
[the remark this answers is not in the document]
plavacquery: no, I will remove lte for date_end and add a lte now+7d for date start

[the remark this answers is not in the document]
plavacquery: in fact yes I will add a better filter on date

Do not take into account my previous remark

idunn/api/utils.py
```diff
@@ -42,6 +42,7 @@
         GradesBlock
     ],
     LIST: [
+        AirQuality,
```
amatissart: why this change ?

idunn/utils/default_settings.yaml
```diff
@@ -75,6 +75,11 @@ BLOCK_AIR_QUALITY_ENABLED: True
 KUZZLE_CLUSTER_URL:
 KUZZLE_REQUEST_TIMEOUT: "0.5" # seconds
 
+# weather API
+WEATHER_API_URL:
+WEATHER_API_KEY:
+WEATHER_REQUEST_TIMEOUT: "0.5" #seconds
+
```
amatissart: this should not be part of this PR

tests/fixtures/api_weather_response.json
```diff
@@ -0,0 +1,15 @@
...
+    "visibility": 10000, 
+    "wind": {"speed": 2.6, "deg": 230}, 
+    "clouds": {"all": 0}, 
+    "dt": 1568044180, 
+    "sys": {"type": 1, "id": 6540, "message": 0.0078, "country": "FR", "sunrise": 1568006291, "sunset": 1568053101},
+    "timezone": 7200, 
+    "id": 2978048,
+    "name": "Saint-Merri",
+    "cod": 200
+}
```
amatissart: this should not be part of this PR

tests/test_events_informations.py
```diff
@@ -72,7 +74,8 @@ def test_event_information_missing_fields():
         type="event_description",
         description="15h-16h [LECTURES D'ALBUMS] Pour les petits (3-6 ans). Accès libre et gratuit.",
         free_text=None,
-        price="Gratuit"
+        price="Gratuit",
+        tags=['']
```
amatissart: I think, we would expect an empty list instead

idunn/api/utils.py
plavacquery: my mistake

idunn/utils/default_settings.yaml
plavacquery: indeed

tests/fixtures/api_weather_response.json
plavacquery: Indeed

idunn/api/places_list.py
```diff
@@ -103,7 +112,15 @@ def valid_category(cls, v):
 
 
 class EventQueryParam(CommonQueryParam):
-    pass
+    outing_category: str = None
```
amatissart: after discussion, to rename to `category`

PR #93, opened 2019-10-09
Directions : Integrate Combigo v1.1

amatissart:
## Description 

**:warning: This PR introduces breaking changes about `RouteLeg` objects for "publictransport" directions.** 

Multiple legs will be returned when multiple transport modes are used successively : all steps in a leg are using the same `"mode"`, that is also present as a new field in the leg object.

As a consequence, the `"stops"` and `"info"` fields are move up to the leg object too.

Geometries objects and properties should be unchanged.


amatissart:
> Looks good to me. I guess we'll see some surprises when requesting from Erdapfel, but nothing obvious to me right now.

Yes, as far as I can see, some content is lost in the roadmap panel with the new version. Probably due to how Erdapfel currently takes the first leg only into consideration.

PR #96, opened 2019-11-29
add weather api

GuillaumeGomez:
Reopening of #88


GuillaumeGomez:
Updated.

idunn/blocks/environment.py
```diff
@@ -66,3 +72,43 @@ def get_air_quality(geobbox):
...
+
+
+def get_local_weather(coord):
+    def inner(coord):
+        return weather_client.fetch_weather_places(coord)
+
+    if not weather_client.enabled:
+        return None
+    key = "{}_{}_{}".format(Weather.BLOCK_TYPE, coord["lat"], coord["lon"])
+    return RedisWrapper.cache_it(key, inner)(coord)
```
jbgriesner: Here the `cache_it()` method will use the default cache timeout, ie 24h, which makes no sense: the weather can change a lot in a 24h timespan.
So we have to define a specific timeout for the weather cache.
I suggest no more than 1h

idunn/blocks/environment.py
```diff
@@ -66,3 +72,43 @@ def get_air_quality(geobbox):
...
+        return cls(**weather)
+
+
+def get_local_weather(coord):
+    def inner(coord):
+        return weather_client.fetch_weather_places(coord)
+
+    if not weather_client.enabled:
+        return None
+    key = "{}_{}_{}".format(Weather.BLOCK_TYPE, coord["lat"], coord["lon"])
```
jbgriesner: also we should probably use approximate coord as key, otherwise the cached values will __probably__ never be accessed.

tests/utils.py
```diff
@@ -42,3 +42,18 @@ def enable_kuzzle():
     """
     with override_settings({"KUZZLE_CLUSTER_URL": "http://localhost:7512"}):
         yield
+
+
+@contextmanager
+def enable_weather_api():
+    """
+    We define here settings specific to tests.
+    We define kuzzle address and port
```
jbgriesner: Is it really related to kuzzle here ?

idunn/blocks/environment.py
GuillaumeGomez: Great catch! I think 30 min would actually be more on point so let's go for this.

idunn/blocks/environment.py
GuillaumeGomez: It's based on cities so it shouldn't be an issue. But just in case, what precision would you use in here?

tests/utils.py
GuillaumeGomez: It isn't anymore, good catch!

idunn/utils/es_wrapper.py
```diff
@@ -8,5 +8,10 @@ def get_elasticsearch():
     global ES_CONNECTION
 
     if ES_CONNECTION is None:
-        ES_CONNECTION = Elasticsearch(settings["MIMIR_ES"])
+        if settings["MIMIR_ES_VERIFY_HTTPS"] is False:
```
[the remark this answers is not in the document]
GuillaumeGomez: I find it a bit less clear actually. I guess it depends on the person...

idunn/utils/redis.py
```diff
@@ -25,3 +29,86 @@ def get_redis_pool(db):
...
+        try:
+            redis_pool = get_redis_pool(db=redis_db)
+        except RedisNotConfigured:
+            logger.warning("No Redis URL has been set for caching", exc_info=True)
+            cls._connection = DISABLED_STATE
+        else:
+            cls._connection = Redis(connection_pool=redis_pool)
+
+    @classmethod
+    def cache_it(cls, key, f, expire=settings["WIKI_CACHE_TIMEOUT"]):
```
amatissart: `expire` is not used if the redis connection is already initialized.
I think it should be passed to `set_value`

PR #97, opened 2019-12-02
Add log message about external direction API call

amatissart:
In order to monitor external calls more easily.

### DIFF
--- a/idunn/directions/client.py
+++ b/idunn/directions/client.py
@@ -105,7 +105,7 @@ class DirectionsClient:
         return DirectionsResponse(**response.json()).dict()
 
 
-    def directions_combigo(self, start, end, mode, lang, **kwargs):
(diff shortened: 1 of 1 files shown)


idunn/directions/client.py
```diff
@@ -159,7 +159,15 @@ def get_directions(self, from_loc, to_loc, mode, lang, **extra):
         else:
             raise BadRequest(f'unknown mode {mode}')
 
-        return method(from_loc, to_loc, mode, lang, **extra)
+        method_name = method.__name__
```
GuillaumeGomez: Why creating a temp var?

amatissart: Because it's used twice in the following call, and it makes easier to replace it with a different name.  
Should I change it ?

GuillaumeGomez: Nope, all good!

PR #98, opened 2019-12-04
Remove apistar_prometheus

GuillaumeGomez:


tests/test_categories.py
```diff
@@ -23,6 +23,7 @@ def test_bbox():
         url=f'http://localhost/v1/places?bbox={BBOX_PARIS}&raw_filter=*,bakery&raw_filter=museum,*&raw_filter=*,place_of_worship'
     )
 
+    print(response.content)
```
jbgriesner: let's remove this print probably :)

PR #99, opened 2019-12-05
api/directions: enable client cache

remi-dupre:
This will help the client not to get blocked by Idunn when using the button to "swap start/end".

NOTE: No other request is currently cached for Idunn, I don't feel like it would be relevant for most of them, except for `places/{id}`, but it still seems of too limited interest.


idunn/utils/default_settings.yaml
```diff
@@ -94,6 +94,7 @@ CORS_OPTIONS_REQUESTS_ENABLED: False
 DIRECTIONS_RL_MAX_REQUESTS: 30 # per client
 DIRECTIONS_RL_EXPIRE: 60 # seconds
 DIRECTIONS_TIMEOUT: 8 # seconds
+DIRECTIONS_CLIENT_CACHE: 600 # seconds
```
amatissart: As a directions response may depend on real-time conditions, this default value may be a little too high.
1 or 2 minutes would be more appropriate, I think.

remi-dupre: I was wondering about it, I thought 10 min was enough compared to the typical duration of a traffic jam, but I guess as we don't really do this for performances... So let's take no risk!

PR #100, opened 2019-12-10
Replace apistar with fastapi

GuillaumeGomez:


app.py
```diff
@@ -1,39 +1,31 @@
...
-    IndexNamesSettingsComponent(),
-    PrometheusComponent()
+    Mount('/v1', routes=get_api_urls(settings)),
 ]
 
-# WARNING: using Classes (and not instances) in hooks list causes a memory leak.
-# See https://github.com/encode/apistar/issues/606 for more details
-event_hooks = [LogErrorHook(), CORSHeaders(), PrometheusHooks()]
-
+app = FastAPI(routes=routes, title="Idunn", debug=True)
```
amatissart: As a reminder, this flag should be enabled for local development only

idunn/api/places.py
```diff
@@ -99,27 +104,28 @@ def get_place(id, es: Elasticsearch, indices: IndexNames, headers: Headers, lang
...
-            'Access-Control-Allow-Methods': 'GET',
-        }
-        return Response('', headers=headers)
-    return Response('', status_code=405)
+        r = Response()
+        response.headers['Access-Control-Allow-Origin'] = '*'
+        response.headers['Access-Control-Allow-Headers'] = request.headers.get('Access-Control-Request-Headers', '*')
+        response.headers['Access-Control-Allow-Methods'] = 'GET'
+    else:
+        response.status_code = 405
```
amatissart: `response` is not defined

idunn/api/places_list.py
```diff
@@ -123,26 +132,38 @@ def valid_categories(cls, v):
...
         )
     return raw_params
 
-def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Settings, query_params: http.QueryParams):
-    raw_params = get_raw_params(query_params)
+
+def get_places_bbox(
+    bbox: Any,
+    category: Optional[str] = Query(None),
+    raw_filter: Optional[str] = Query(None),
```
amatissart: Is this correct, given that this route should accept multiple `category` or `raw_filter` in the same query ?

idunn/api/places_list.py
```diff
@@ -151,7 +172,7 @@ def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Sett
             # PJ is currently the only source that accepts arbitrary queries
             source = SOURCE_PAGESJAUNES
         elif params.category \
-            and all(c.get('pj_filters') for c in params.category) \
+            and params.category.get('pj_filters') \
```
amatissart: what if multiple categories are present in the query ?

idunn/api/places_list.py
```diff
@@ -182,16 +203,20 @@ def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Sett
...
     if not kuzzle_client.enabled:
-        raise HTTPException("Kuzzle client is not available", status_code=501)
+        raise HTTPException(
+            status_code=501,
+            detail="Kuzzle client is not available"
+        )
 
     try:
-        params = EventQueryParam(**query_params)
+        params = EventQueryParam(**{'category': category, 'bbox': bbox})
```
amatissart: some params are missing, such as `size`, `lang` and `verbosity`

idunn/blocks/base.py
```diff
@@ -50,9 +53,10 @@ def validate(self, value, definitions=None, allow_coerce=False):
 
         return block_class.validate(value=value, definitions=definitions,
                                     allow_coerce=allow_coerce)
+        return value
```
amatissart: `return` twice ?

idunn/blocks/happy_hour.py
```diff
@@ -13,22 +14,22 @@
...
-
-    status = validators.String(enum=STATUSES)
-    next_transition_datetime = validators.String(allow_null=True)
-    seconds_before_next_transition = validators.Integer(allow_null=True)
-    raw = validators.String()
-    days = validators.Array(items=DaysType)
+    BLOCK_TYPE: ClassVar = 'happy_hours'
+    STATUSES: ClassVar = ['yes', 'no']
+
+    status: constr(regex='({})'.format('|'.join(STATUSES)))
```
amatissart: Maybe it would be cleaner to define an Enum ?  
https://pydantic-docs.helpmanual.io/usage/types/#enums-and-choices

idunn/utils/logging.py
```diff
@@ -32,8 +32,9 @@ def init_logging(settings: Settings):
     logging.getLogger().handlers = [logHandler]
 
 
-class LogErrorHook:
-    def on_error(self, request: http.Request):
-        prometheus.exception("unhandled_error")
-        logging.getLogger('idunn.error')\
-            .exception("An unhandled error was raised.", extra={'url': request.url})
+async def handle_errors(request: Request, exception):
+    print('hello!')
```
amatissart: to clean up

idunn/utils/rate_limiter.py
```diff
@@ -10,14 +10,17 @@
...
 
+
 @contextmanager
 def dummy_limit():
     yield
 
+
 class HTTPTooManyRequests(HTTPException):
     default_status_code = 429
     default_detail = 'Too Many Requests'
```
amatissart: this derived class may not work with FastAPI

idunn/api/closest.py
```diff
@@ -23,16 +24,20 @@ def get_closest_place(lat: float, lon: float, es: Elasticsearch):
 
     if loader is None:
         prometheus.exception("FoundPlaceWithWrongType")
-        raise Exception("Closest address to '{}:{}' has a wrong type: '{}'".format(lat, lon, es_addr.get('_type')))
+        raise HTTPException(
+            status_code=404,
+            detail="Closest address to '{}:{}' has a wrong type: '{}'".format(lat, lon, es_addr.get('_type')))
```
amatissart: As this error would show a problem in the database, it should not fail silently.  
We should keep a warning message at least.

idunn/api/directions.py
```diff
@@ -11,18 +12,19 @@
...
 
     from_position = (f_lon, f_lat)
     to_position = (t_lon, t_lat)
-    params_dict = dict(params)
-    mode = params_dict.pop('type', '')
-    lang = params_dict.pop('language', '') or 'en'
 
     if not mode:
-        raise BadRequest('"type" query param is required')
+        raise HTTPException(status_code=401, detail='"type" query param is required')
```
amatissart: status_code should be 400 here

idunn/api/places_list.py
```diff
@@ -182,16 +207,20 @@ def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Sett
     }
 
 
-def get_events_bbox(bbox, query_params: http.QueryParams):
+def get_events_bbox(bbox: str, category: str, size: int = Query(None)):
```
amatissart: `category` query param should not be required. This may explain why the kuzzle test is failing.

app.py
```diff
@@ -1,39 +1,53 @@
...
-    PrometheusComponent()
-]
+app = FastAPI(routes=routes, title="Idunn", debug=True)
+
+
+@app.middleware("http")
+async def db_session_middleware(request: Request, call_next):
+    response = await call_next(request)
+    response.headers['Access-Control-Allow-Origin'] = '*'
+    return response
```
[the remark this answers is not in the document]
GuillaumeGomez: I'll at least add a TODO otherwise we might forget...

app.py
```diff
@@ -1,39 +1,53 @@
...
+    Mount('/v1', routes=get_api_urls(settings)),
 ]
 
-components = [
-    settings,
-    ElasticSearchComponent(),
-    IndexNamesSettingsComponent(),
-    PrometheusComponent()
-]
+app = FastAPI(routes=routes, title="Idunn", debug=True)
```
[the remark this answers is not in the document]
GuillaumeGomez: I'll add it only if launched with `__name__ == "main"`.

idunn/api/closest.py
GuillaumeGomez: Isn't it why we have `prometheus.exception("FoundPlaceWithWrongType")`?

idunn/api/places_list.py
```diff
@@ -151,7 +176,7 @@ def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Sett
             # PJ is currently the only source that accepts arbitrary queries
             source = SOURCE_PAGESJAUNES
         elif params.category \
-            and all(c.get('pj_filters') for c in params.category) \
+            and params.category.get('pj_filters') \
```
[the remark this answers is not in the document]
GuillaumeGomez: Great point and good catch. Should we create a ticket for this test issue?

idunn/blocks/contact.py
```diff
@@ -1,17 +1,21 @@
...
 
     @classmethod
     def from_es(cls, es_poi, lang):
+        print(es_poi.properties)
         mail = es_poi.properties.get('email') or es_poi.properties.get('contact:email')
         if not mail:
             return None
+        # TODO: maybe reuse pydantic for this check?
+        if not isinstance(mail, str):
+            return None
```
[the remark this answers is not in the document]
GuillaumeGomez: From my point of view it is.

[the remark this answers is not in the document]
GuillaumeGomez: (problem here is that pydantic can't be used directly since `mail` isn't a field...)

idunn/blocks/happy_hour.py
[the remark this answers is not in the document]
GuillaumeGomez: Oh nice! Thanks!

idunn/api/places_list.py
GuillaumeGomez: Yep it was that. Good catch once again!

idunn/directions/models.py
```diff
@@ -159,7 +159,7 @@ class RoutePrice(BaseModel):
 class DirectionsRoute(BaseModel):
     duration: int = Schema(..., description='duration in seconds')
     distance: Optional[int] = Schema(..., description='distance in meters')
-    carbon: Optional[float] = Schema(..., description='value in gEC')
+    carbon: Optional[float] = None # Schema(..., description='value in gEC')
```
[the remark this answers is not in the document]
GuillaumeGomez: If not, I get this error (and I didn't check further):

```
pydantic.error_wrappers.ValidationError: 3 validation errors for DirectionsResponse
data -> routes -> 0 -> carbon
  field required (type=value_error.missing)
data -> routes -> 1 -> carbon
  field required (type=value_error.missing)
data -> routes -> 2 -> carbon
  field required (type=value_error.missing)
```

idunn/utils/app.py
```diff
@@ -1,6 +1,16 @@
-from apistar import App
+from fastapi import FastAPI
```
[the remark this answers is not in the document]
GuillaumeGomez: Indeed, deleted!

idunn/utils/rate_limiter.py
GuillaumeGomez: Is it tested somewhere?

idunn/api/closest.py
amatissart: Prometheus would not give any detail about the query that caused the error. It's only used to build metrics.

In this case where the error is quite unexpected and may bring another issue to our attention, an error message is much more useful than a counter.

tests/test_kuzzle.py
```diff
@@ -28,46 +28,46 @@ def test_kuzzle_event_ok():
...
-            }
-        ]
-
-
-        """
-        We check that there is no Wikipedia block
-        in the answer
-        """
-        assert all(b['type'] != "wikipedia" for b in firstEventData['blocks'])
+        # TODO (none found, might be because of the mock)
```
amatissart: Is there still an issue about this kuzzle test ?

idunn/utils/rate_limiter.py
amatissart: I will add a test for this, in a separate PR.

amatissart: A new test case is defined in https://github.com/QwantResearch/idunn/pull/102 to test this behavior

idunn/api/places_list.py
```diff
@@ -78,43 +80,53 @@ def valid_bbox(cls, v):
...
 class PlacesQueryParam(CommonQueryParam):
-    category: List[Any] = []
-    raw_filter: List[str] = None
-    source: Optional[str] = None
-    q: Optional[str] = None
+    category: List[Any]
+    raw_filter: Optional[List[str]]
+    source: Optional[str]
+    q: Optional[str]
+    size: Optional[int]
```
amatissart: `size` is already defined in `CommonQueryParam`

idunn/api/places_list.py
```diff
@@ -123,26 +135,40 @@ def valid_categories(cls, v):
...
-def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Settings, query_params: http.QueryParams):
-    raw_params = get_raw_params(query_params)
+
+def get_places_bbox(
+    bbox: Any,
+    category: Optional[List[str]] = Query(None),
+    raw_filter: Optional[List[str]] = Query(None),
+    source: Optional[str] = Query(None),
+    q: Optional[str] = Query(None),
+    size: Optional[int] = Query(None)
```
amatissart: `lang` and `verbosity` should also be read from the query params

idunn/api/places_list.py
```diff
@@ -182,16 +208,20 @@ def get_places_bbox(bbox, es: Elasticsearch, indices: IndexNames, settings: Sett
...
     if not kuzzle_client.enabled:
-        raise HTTPException("Kuzzle client is not available", status_code=501)
+        raise HTTPException(
+            status_code=501,
+            detail="Kuzzle client is not available"
+        )
 
     try:
-        params = EventQueryParam(**query_params)
+        params = EventQueryParam(**{'category': category, 'bbox': bbox, 'size': size})
```
amatissart: `lang` and `verbosity` are missing here too

idunn/utils/rate_limiter.py
GuillaumeGomez: Thanks!

idunn/directions/models.py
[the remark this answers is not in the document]
amatissart: Then, I suppose that an explicit default value needs to be set:
```suggestion
    carbon: Optional[float] = Schema(None, description='value in gEC')
```

Dockerfile
```diff
@@ -29,5 +29,5 @@ ADD --chown=app_user gunicorn_logging.conf .
 
 # You can set the number of workers by passing --workers=${NB_WORKER} to the docker run command.
 # For some reason, an array is required here to accept other params on run.
-ENTRYPOINT ["gunicorn", "app:app", "--bind=0.0.0.0:5000", "--pid=pid", \
-  "--worker-class=meinheld.gmeinheld.MeinheldWorker", "--preload", "--log-config=/app/gunicorn_logging.conf"]
+ENTRYPOINT ["uvicorn", "app:app", "--host=0.0.0.0", "--port=5000", \
+  "--log-config=/app/gunicorn_logging.conf"]
```
amatissart: In a production environment, uvicorn documentation recommends to keep gunicorn, using `UvicornWorker` as the worker class. https://www.uvicorn.org/deployment/#gunicorn

This also makes possible to keep our current deployment and logging configurations, to use multiple workers by using `--workers` in the docker command, and to keep the `--preload` option to reduce the memory usage.

tests/test_kuzzle.py
GuillaumeGomez: Yep, when running it there is nothing returned.

GuillaumeGomez: Yep, when running it there is nothing returned.

GuillaumeGomez:
Fixed the dependencies.

PR #101, opened 2019-12-31
restrict public transports locations

remi-dupre:
**:warning: This pull request is based on https://github.com/QwantResearch/idunn/pull/100/files**

Only allow public transport requests inside the neighborhood of Paris or Lyon. If a requested route isn't inside one of these two zones, an error is raised with code 400.


idunn/utils/default_settings.yaml
```diff
@@ -91,6 +91,7 @@ CORS_OPTIONS_REQUESTS_ENABLED: False
 
 #######################
 ## Directions
+DIRECTIONS_ALLOWED_CITIES: ['paris', 'lyon']
```
GuillaumeGomez: So we limit this to only a few cities in the end? I thought it was only for public transports though. Did I miss something?

remi-dupre: No you're right, my settings doesn't have a correct name!

idunn/api/directions.py
```diff
@@ -40,7 +40,7 @@ def get_directions(
 
         if not allowed_zone:
             raise HTTPException(
-                status_code=403,
+                status_code=406,
```
bbecquet: I don't think that's the right status to use, 406 is about content negotiation, using the `Accept-*` HTTP headers sent by the client to determine if the server can produce a content matching the criteria (language, MIME type, etc.) https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

remi-dupre: The final choice seems to be 422.

idunn/utils/geometry.py
```diff
@@ -55,9 +62,21 @@ def parse_poly(lines):
...
 
 
+# Load shape for France
 with open(france_poly_filename) as france_file:
     france_polygon = parse_poly(france_file.readlines())
 
+# Load shape for some cities surrounding
+city_surrounds_polygons = dict()
+
+for filename in os.listdir(cities_surrounds_path):
```
bbecquet: An alternative would be to use in a single geojson file with a feature for each city… It would simplify this file handling and need less config, but not sure it would be more flexible. What do you think?

idunn/api/directions.py
```diff
@@ -15,25 +20,35 @@
...
 ):
     rate_limiter.check_limit_per_client(request)
 
     from_position = (f_lon, f_lat)
     to_position = (t_lon, t_lat)
 
-    if not type:
-        raise HTTPException(status_code=400, detail='"type" query param is required')
+    if type in ("publictransport", "taxi", "vtc", "carpool"):
+        allowed_zone = any(
```
bbecquet: Maybe we can put this logic in a separate function, even in the `geometry.py` file itself, so it can be maintained and tested separately.

remi-dupre: I totally agree, just moved it here: https://github.com/QwantResearch/idunn/pull/101/files#diff-11182b086324030b6623bf1b6cdcce50R40-R51

idunn/utils/geometry.py
amatissart: I agree that a single file may be easier to handle.  
The config is still useful to be able to allow/disallow some cities on deploy, though.

amatissart:
There are also minor conflicts to solve, following #109 
cc @remi-dupre

PR #103, opened 2020-01-06
Fix kuzzle tests with FastAPI

amatissart:
Besides tweaking the parameters validation, it is also necessary to define a custom formatter for datetimes, in order to preserve the same format.


GuillaumeGomez:
Thanks!

PR #105, opened 2020-01-07
add autocomplete route for bragi

remi-dupre:
Add a route for auto-completion. This is currently almost equivalent to proxying Bragi, however this will get useful in the future to analyze the request on top of the geocoder.

**:warning: there are a few changes to the API compared to bragi:**

 - `_debug` parameter as been removed
 - list parameters are not suffixed with `[]`, for example use `?type=poi&type=house` instead of bragi syntax: `?type[]=poi&type[]=house`


idunn/api/geocoder.py
```diff
@@ -0,0 +1,20 @@
...
+from pydantic import BaseModel
+
+from idunn import settings
+from ..geocoder.client import geocoder_client
+
+
+class GeocoderParams(BaseModel):
+    shape: dict = Query({}, title='restrict search inside of a polygon')
+
+def get_autocomplete(
```
GuillaumeGomez: Please add an empty line before this function (elements must be separated by two empty lines according to PEP).

idunn/geocoder/client.py
```diff
@@ -0,0 +1,31 @@
...
+            params.update({
+                'lon': lon,
+                'lat': lat,
+            })
+
+        response = self.session.get(
+            settings['BRAGI_BASE_URL'] + '/autocomplete',
+            params=params
+        )
+        response.raise_for_status() # TODO: catch errors in [400, 500[
```
GuillaumeGomez: "500\["? ;)

GuillaumeGomez:
We just need a test to confirm the expected behavior (that it's redirecting) but otherwise looks good to me.

idunn/geocoder/models.py
```diff
@@ -0,0 +1,177 @@
...
+class GeocodingResponse(BaseModel):
+    type: str
+    label: Optional[str]
+    name: Optional[str]
+    housenumber: Optional[str]
+    street: Optional[str]
+    locality: Optional[str]
+    postcode: Optional[str]
+    city: Optional[str]
+    #  accuracy: Optional[PositiveInt]
```
GuillaumeGomez: Please remove those fields if they're not used.

idunn/geocoder/models.py
```diff
@@ -0,0 +1,177 @@
...
+    comments: List[Comment] = []
+    physical_modes: List[PhysicalMode] = []
+    lines: List[Line] = []
+    timezone: Optional[str]
+    codes: List[Code] = []
+    feed_publishers: List[FeedPublished] = []
+    bbox: Optional[Rect]
+    country_codes: List[str] = []
+
+GeocodingResponse.update_forward_refs()
```
GuillaumeGomez: Please add an empty line before this one.

idunn/geocoder/models.py
```diff
@@ -0,0 +1,177 @@
...
+
+class FeatureProperties(BaseModel):
+    geocoding: GeocodingResponse
+
+
+class Explaination(BaseModel):
+    value: float
+    description: str
+    details: List['Explaination']
+
```
GuillaumeGomez: Please add an empty line before this one (PEP is back!)

idunn/geocoder/models.py
```diff
@@ -0,0 +1,177 @@
...
+
+class FeatureProperties(BaseModel):
+    geocoding: GeocodingResponse
+
+
+class Explaination(BaseModel):
+    value: float
+    description: str
+    details: List['Explaination']
```
GuillaumeGomez: Isn't it weird to have a type defined as a string?

idunn/geocoder/models.py
remi-dupre: The thing is that they are in the "specification" (https://github.com/geocoders/geocodejson-spec), but they are commented out in Bragi: https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/model.rs#L194-L199.

I should have at least added a comment about that but couldn't it be relevant to keep them here in order to "explain what is missing from the standard" while avoiding using them when Bragi doesn't ?

GuillaumeGomez: Hum... Then add a comment over the commented fields so people don't wonder why they are still here.

idunn/geocoder/models.py
remi-dupre: That is kind of weird, but this is the best way I found to fix circular dependancies in the models. 

I just discovered that since Python 3.7 it seems that it is possible to [delay the parsing of type annotations to runtime](https://www.python.org/dev/peps/pep-0563/#resolving-type-hints-at-runtime), however it appears that Idunn should still be 3.6-compatible.

idunn/geocoder/client.py
```diff
@@ -0,0 +1,55 @@
...
+        if shape:
+            response = self.session.post(url, params=params, json={"shape": shape})
+        else:
+            response = self.session.get(url, params=params)
+
+        if response.status_code != requests.codes.ok:
+            try:
+                explain = response.json()["long"]
+            except (IndexError, JSONDecodeError):
+                explain = "unknown reason"
```
amatissart: Including the content of the response as plain text may be useful, using `response.text`

idunn/geocoder/client.py
```diff
@@ -0,0 +1,55 @@
...
+
+            logger.error(
+                'Request to Bragi returned with unexpected status %d: "%s"',
+                response.status_code,
+                explain,
+            )
+            raise HTTPException(500)
```
amatissart: A short message with the error could be helpful, such as "Unexpected geocoder error".  
Also, I think 503 would be more appropriate, as this error is not directly caused by Idunn.

idunn/utils/default_settings.yaml
```diff
@@ -100,3 +100,8 @@ MAPBOX_DIRECTIONS_API_BASE_URL: "https://api.mapbox.com/directions/v5/mapbox"
 MAPBOX_DIRECTIONS_ACCESS_TOKEN:
 COMBIGO_API_BASE_URL: "https://maps.combigo.com/v1.1"
 COMBIGO_API_KEY:
+
+#######################
+## Geocoding
+
+BRAGI_BASE_URL: "http://localhost:4000"
```
amatissart: `http://bragi:4000` could be a better default, both for testing and deployment

idunn/api/geocoder.py
```diff
@@ -0,0 +1,23 @@
...
+    shape: dict = Query(None, title="restrict search inside of a polygon")
+
+
+def get_autocomplete(
+    extra: ExtraParams = ExtraParams(),
+    query: str = Query(..., alias="q", title="query string"),
+    lon: Optional[float] = Query(None, ge=-180, le=180, title="latitude for the focus"),
+    lat: Optional[float] = Query(None, ge=-90, le=90, title="longitude for the focus"),
+    lang: str = Query(settings["DEFAULT_LANGUAGE"], title="language"),
+    limit: int = Query(10, ge=1, title="maximum number of results"),
```
amatissart: A few useful parameters are missing here: `offset`, `type[]`, `zone_type[]`, `poi_type[]`  
see https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/routes/autocomplete.rs#L59

remi-dupre: I feel ashamed for this one, I just expected to add them when I was done with the core structure... :(

remi-dupre: I added missing fields here: https://github.com/QwantResearch/idunn/pull/105/commits/63b9f079e87279dab8cf6d748dc6c0fe0c1f120f#diff-853016d15221f48bae06f82ba029d5ae

I had a hard time trying to elegantly get a documentation of the parameters while keeping them packed in a pedantic Model (which is quite convenient to use) since it requires to anotate the constructor.

idunn/geocoder/models/params.py
```diff
@@ -0,0 +1,87 @@
...
+        should be sent to bragi.
+        """
+        return {
+            (key if not isinstance(value, list) else key + "[]"): value
+            for (key, value) in self.dict().items()
+        }
+
+
+class ExtraParams(BaseModel):
+    shape: dict = Query(None, title="restrict search inside of a polygon")
```
amatissart: should that not be a `Body()` value ?

idunn/geocoder/models/params.py
```diff
@@ -0,0 +1,87 @@
...
+class Type(str, Enum):
+    City = "city"
+    House = "house"
+    Poi = "poi"
+    StopArea = "public_transport:stop_area"
+    Street = "street"
+    Zone = "zone"
+
+
+class QueryParams(BaseModel):
```
amatissart: It is actually possible to simplify this class and keep the generated docs, thanks to Pydantic's `dataclass` decorator:  
https://pydantic-docs.helpmanual.io/usage/dataclasses/

idunn/api/geocoder.py
```diff
@@ -0,0 +1,14 @@
+from typing import List, Optional
+
+from fastapi import Query, Depends
+from pydantic import BaseModel, PositiveInt
+
+from idunn import settings
```
amatissart: most of these imports are not used

idunn/geocoder/client.py
```diff
@@ -0,0 +1,41 @@
...
+                explain = response.text
+
+            logger.error(
+                'Request to Bragi returned with unexpected status %d: "%s"',
+                response.status_code,
+                explain,
+            )
+            raise HTTPException(503, "Unexpected geocoder error")
+
+        return GeocodeJson.parse_obj(response.json())
```
GuillaumeGomez: Shouldn't we put this into a try block?

idunn/geocoder/models/geocodejson.py
```diff
@@ -0,0 +1,170 @@
+"""
+Implement GeocodeJson specification as defined here:
+ - https://github.com/geocoders/geocodejson-spec/tree/master/draft
+ - https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/model.rs
+"""
+from typing import List, Optional, Tuple
+
+from pydantic import BaseModel, PositiveInt, confloat
+
```
GuillaumeGomez: Do we have to put an empty between each import? :)

remi-dupre: I usually respect this convention:

 - standard library imports
 - (empty line)
 - external libraries imports
 - (empty line)
 - project level imports

I'm not sure this is standard but most people seem to follow something close to that? I can remove this empty line if you prefer!

GuillaumeGomez: This is just a nit so I don't really attach much importance to it. If the formatter is happy, then so am I.

idunn/geocoder/client.py
remi-dupre: Just pushed https://github.com/QwantResearch/idunn/pull/105/commits/50d537e34a7c16afc2cfc8177c7440661f80a55e

PR #106, opened 2020-01-08
Make logging and prometheus metrics more consistent with previous behavior

amatissart:
* Replace starlette-prometheus with custom `MonitoredAPIRoute` to keep detailed metrics (with `handler` label) that were defined in apistar-prometheus

* Simplify logging configuration: a unique configuration is defined in utils/logging.py and gunicorn configuration is longer necessary


app.py
```diff
@@ -2,27 +2,21 @@
...
-    Mount('/v1', routes=get_api_urls(settings)),
-]
-
-app = FastAPI(routes=routes, title="Idunn", debug=__name__ == '__main__')
-
+v1_routes = get_api_urls(settings)
+app.include_router(
+    APIRouter(v1_routes),
+    prefix='/v1'
+)
```
amatissart: Thanks to FastAPI `APIRouter`, all routes are now present in the openapi schema, and in the generated docs

PR #107, opened 2020-01-10
Black

GuillaumeGomez:


GuillaumeGomez:
> It may be usefull to add a pyproject.toml, otherwise `black` might use user-specific configurations: https://black.readthedocs.io/en/stable/pyproject_toml.html#where-black-looks-for-the-file

Sounds fine to me as is. Let's see what others think about it.

> Do we want to keep default settings ? I personal don't mind with 88 columns and we don't use any string convention between single or double quote in our codebase. 

I personally like it as is so if it's fine with everyone, let's keep it that way.

> Maybe it would be safer to specify that we use Python 3.6.

It doesn't work with previous python versions apparently so I guess we're good.

.travis.yml
```diff
@@ -10,10 +10,11 @@ env:
   - PIPENV_IGNORE_VIRTUALENVS=1
 
 install:
-  - pip install pipenv
+  - pip install pipenv black
   - pipenv install --dev --deploy
 
 script:
+  - black --diff --check idunn/ tests/
```
remi-dupre: It could help to list directories we want to format in the pyproject.toml, we would only need to run `black .` and it will be easier to use with an IDE.

remi-dupre:
> It doesn't work with previous python versions apparently so I guess we're good.

Oh, fine, there is still a way to specify it, I'm not sure what is the difference though: https://github.com/prologin/django-proloauth-client/blob/c39f3c8e2693a94c1b0aa0103a2f99cea3cb9731/pyproject.toml#L3

.travis.yml
```diff
@@ -10,10 +10,11 @@ env:
   - PIPENV_IGNORE_VIRTUALENVS=1
 
 install:
-  - pip install pipenv
+  - pip install pipenv black
```
amatissart: I would prefer to add `black` to dev dependencies in Pipfile

GuillaumeGomez: I tried, we can't. There are only pre-releases available. pipenv doesn't like it much.

amatissart: `pipenv install --dev "black==19.10b0"` does work for me

GuillaumeGomez: Interesting! Let's try again then.

GuillaumeGomez:
Fixed comments. I also changed the column limit to 100. It's more common nowadays.

.travis.yml
```diff
@@ -14,6 +14,7 @@ install:
   - pipenv install --dev --deploy
 
 script:
+  - black --diff --check .
```
amatissart:
```suggestion
  - pipenv run black --diff --check .
```

GuillaumeGomez:
Updated.

pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.black]
+line-length = 100
```
remi-dupre: The default value seems to be 88, it appears that it was totally overthought: https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length

I don't mind if we change it to `<= 100` (but maybe we don't want to change it if nobody has an opinion about it).

GuillaumeGomez: Generally, you either see 80 or 100 so I'm surprised with your choice of 99. :)

remi-dupre: \* "their choice of 88" ?

> In general, 90-ish seems like the wise choice.

I agree that's a funny choice, their justification appears to be "88 is 10% more than 80" :D

GuillaumeGomez:
Well to the "black" age! \o/

PR #108, opened 2020-01-10
fix trailing commas added by black

remi-dupre:
As black was run with different settings several times, some strange behaviors appeared with trailing commas. This is a bit prettier like this plus it will be a bit easier for me to rebase with only one run to apply.

### DIFF
--- a/idunn/api/kuzzle.py
+++ b/idunn/api/kuzzle.py
@@ -114,7 +114,7 @@ class KuzzleClient:
                                 "multi_match": {
                                     "query": category,
                                     "type": "best_fields",
-                                    "fields": ["tag^5", "free_text^4", "title^4", "description^3",],
+                                    "fields": ["tag^5", "free_text^4", "title^4", "description^3"],
                                     "tie_breaker": 0.7,
                                 },
                             }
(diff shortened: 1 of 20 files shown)


GuillaumeGomez:
Can we enforce it in the black setting file as well?

remi-dupre:
> Can we enforce it in the black setting file as well?

I don't think so, that's not really a fix on current code but just a cleaner application of black to master~1. (Current master is considered Ok by black, maybe the trailing comma can matter in python 3.6 ?)

remi-dupre:
Accordant to [the documentation](https://test-black.readthedocs.io/en/style-guide/style_guide/trailing_commas.html):

> One exception to adding trailing commas is function signatures containing *, *args, or **kwargs. In this case a trailing comma is only safe to use on Python 3.6. Black will detect if your file is already 3.6+ only and use trailing commas in this situation. If you wonder how it knows, it looks for f-strings and existing use of trailing commas in function signatures that have stars in them. In other words, if you’d like a trailing comma in this situation and Black didn’t recognize it was safe to do so, put it there manually and Black will keep it.

PR #112, opened 2020-01-17
New directions endpoint with place id parameters

amatissart:
### A new endpoint for directions

**GET**  `/v1/directions?origin=osm:node:4242&destination=latlon:2.9:12.3&type=publictransport`
where `origin` and `destination` are place ids.

It is now possible for the server to fetch details about the places used as origin and destination, and implement specific behavior for public transports stops.

This comes in addition to the existing route
**GET** `/v1/directions/{f_lon},{f_lat};{t_lon},{t_lat}`  
(with explicit latitudes and longitudes in the URL path)

Except the new `origin` and `destination`, the new route accepts the same query parameters (`type`, `language`, etc.)

----

#### A note about Elasticsearch indices
By default, Idunn will now fetch POIs from both `munin` index (shared with the geocoder) and `munin_poi_nosearch` (introduced by https://github.com/QwantResearch/fafnir/pull/40) containing extra POIs that are not included in the search results.

----

#### `place_from_id`

This PR introduces a new utils function `place_from_id` to fetch details about any kind of place id used by Qwant Maps.

### DIFF
--- a/idunn/api/directions.py
+++ b/idunn/api/directions.py
@@ -1,14 +1,12 @@
-from fastapi import HTTPException
-
-from pydantic import constr
-
+from fastapi import HTTPException, Query, Depends
 from starlette.requests import Request
 from starlette.responses import Response
 
 from idunn import settings
-from idunn.utils.geometry import city_surrounds_polygons
+from idunn.places import Latlon, place_from_id, InvalidPlaceId
 from idunn.utils.rate_limiter import IdunnRateLimiter
 from ..directions.client import directions_client
+
 
 rate_limiter = IdunnRateLimiter(
     resource="idunn.api.directions",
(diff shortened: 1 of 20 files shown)


idunn/api/directions.py
```diff
@@ -1,37 +1,68 @@
...
-    type: constr(min_length=1),
-    language: str = "en",  # query parameters
+    t_lat: float,
+    # Query parameters
+    type: str,
+    language: str = "en",
+    # Request
+    request: Request = Depends(directions_request),
+):
+
```
GuillaumeGomez: Why the empty line?

idunn/api/urls.py
```diff
@@ -33,19 +33,27 @@ def get_api_urls(settings):
     return [
         APIRoute("/metrics", metric_handler),
         APIRoute("/status", get_status),
-        # Deprecated
-        APIRoute("/pois/{id}", get_poi),
+        APIRoute("/pois/{id}", get_poi, deprecated=True),
```
GuillaumeGomez: Keeping the comment might be a good idea I think.

PR #118, opened 2020-02-19
Upgrade from Python 3.6 to Python 3.8

amatissart:
The main motivation for this change is to limit the number of threads used by FastAPI.

FastAPI + Starlette use the [`ThreadPoolExecutor`](https://docs.python.org/fr/3.8/library/concurrent.futures.html?highlight=concurrent#concurrent.futures.ThreadPoolExecutor) internally to handle non-async routes.
In Python <3.8, the maximum number of workers used by this executor is `os.cpu_count() * 5`.
That is unnecessary high on machines with >=32 cpus, where multiple instances of Idunn may be launched.
See https://github.com/tiangolo/fastapi/issues/603 for more details.

In Python 3.8, the number of workers is bounded to 32, which is more reasonable.

We could workaround this default value differently, but Python 3.6 also lacks some features that could be handy in the future, such as [Postponed Evaluation of Annotations](https://pydantic-docs.helpmanual.io/usage/postponed_annotations/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,14 @@ The main endpoints are:
 
 ## Running
 
-- The dependencies are managed with [Pipenv](https://github.com/pypa/pipenv).
-- To run the api you need to do:
+### Requirements
+
+   - Python 3.8
+   - [Pipenv](https://github.com/pypa/pipenv), to manage dependencies and virtualenv
+
+### Installation
+
+- Create the virtualenv and install dependencies:
   ```shell
   pipenv install
   ```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -43,8 +43,15 @@ The main endpoints are:
...
 ## Running
 
-- The dependencies are managed with [Pipenv](https://github.com/pypa/pipenv).
-- To run the api you need to do:
+### Requirements
+
+   - Python 3.8
+   - [Pipenv](https://github.com/pypa/pipenv), to manage dependencies and virtualenv
+
+
```
GuillaumeGomez: Please remove this empty line.

PR #119, opened 2020-02-20
Define settings to enable /docs endpoint with custom path prefix

amatissart:
Define 2 new settings values:
 * `DOCS_ENABLED`: to enable `/docs` endpoint provided by FastAPI (disabled by default)
 * `DOCS_URL_PREFIX`: the base path used to access the OpenAPI schema 
(required when Idunn is deployed behind a reversed proxy with a path different from `/`) 

image

### DIFF
--- a/app.py
+++ b/app.py
@@ -11,12 +11,10 @@ import uvicorn
 
 # Setup docs settings
 docs_settings = {
-    "docs_url": None,
-    "redoc_url": "/docs",
     "openapi_prefix": settings["DOCS_URL_PREFIX"],
 }
 if not settings["DOCS_ENABLED"]:
-    docs_settings.update({"openapi_url": None, "redoc_url": None})
+    docs_settings.update({"openapi_url": None, "redoc_url": None, "docs_url": None})
 
 
 # Setup FastAPI app
--- a/app.py
+++ b/app.py
@@ -9,29 +9,38 @@ from starlette.responses import PlainTextResponse
 import uvicorn
 
 
-app = FastAPI(title="Idunn", debug=__name__ == '__main__')
+# Setup docs settings
+docs_settings = {
+    "docs_url": None,
+    "redoc_url": "/docs",
+    "openapi_prefix": settings["DOCS_URL_PREFIX"],
+}
+if not settings["DOCS_ENABLED"]:
+    docs_settings.update({"openapi_url": None, "redoc_url": None})
 
+
+# Setup FastAPI app
+app = FastAPI(title="Idunn", version="0.2", debug=__name__ == "__main__", **docs_settings)
 v1_routes = get_api_urls(settings)
-app.include_router(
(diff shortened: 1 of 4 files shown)


idunn/api/urls.py
```diff
@@ -22,23 +22,20 @@ def get_metric_handler(settings):
...
-        APIRoute("/status", get_status),
+        APIRoute("/metrics", metric_handler, include_in_schema=False),
+        APIRoute("/status", get_status, include_in_schema=False),
         # Deprecated POI route
         APIRoute("/pois/{id}", get_poi, deprecated=True),
         # Places
         APIRoute("/places", get_places_bbox),
         APIRoute("/places/latlon:{lat}:{lon}", get_place_latlon),
-        APIRoute("/places/{id}", handle_option, methods=["OPTIONS"]),
+        APIRoute("/places/{id}", handle_option, methods=["OPTIONS"], include_in_schema=False),
```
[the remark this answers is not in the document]
amatissart: They are internal endpoints, used for monitoring in particular. So application developers should not rely on them.

app.py
```diff
@@ -9,29 +9,38 @@
 import uvicorn
 
 
-app = FastAPI(title="Idunn", debug=__name__ == '__main__')
-
+# Setup docs settings
+docs_settings = {
+    "docs_url": None,
```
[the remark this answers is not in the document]
amatissart: That's a good point. I find that redoc is a bit more convenient to browse into the documentation.
I suppose we can simply keep both of them.

app.py
```diff
@@ -11,12 +11,10 @@
 
 # Setup docs settings
 docs_settings = {
-    "docs_url": None,
-    "redoc_url": "/docs",
     "openapi_prefix": settings["DOCS_URL_PREFIX"],
```
remi-dupre: I'm not sure this settings is useful, I fell like it is used to customize where the docs will look for `openapi.json` ([which will be served at the root by default for fastapi](https://fastapi.tiangolo.com/advanced/extending-openapi/#the-normal-process)).

Maybe it also have a wrong name, as I understand it it may not affect the url of documentation endpoints?

amatissart: Swagger and Redoc clients will indeed try to fetch the schema via `/openapi.json` (using an absolute path). 
But Idunn is typically deployed under `/maps/detail` prefix. And the clients would fail to fetch the schema file using this absolute path.
With `DOCS_URL_PREFIX=/maps/detail`, docs will be available via `/maps/detail/docs` and the clients will fetch `openapi.json` correctly from `/maps/detail/openapi.json`

Do you thing this setting name should be changed to be more explicit ?

remi-dupre: Oh no you're right, I assumed it would use a relative path, this makes more sense.

About the name I'm not sure anything I could think of would really be better, I guess this is fine :)

PR #120, opened 2020-03-04
Use httpx to make bragi and nlu requests concurrently

amatissart:
This PR defines an `async def`  function to handle the /autocomplete endpoint.  

Using [`httpx`](https://www.python-httpx.org) (as an alternative to `requests`) and its **`AsyncClient`**, it is possible to fetch bragi and nlu responses concurrently, and reduce latency.

For tests, [`respx`](https://lundberg.github.io/respx/) is used as an alternative to `responses` to mock http responses fetched by the httpx client.


idunn/geocoder/bragi_client.py
```diff
@@ -1,38 +1,35 @@
...
+        self.client = httpx.AsyncClient(verify=settings["VERIFY_HTTPS"])
 
-    def autocomplete(self, query: QueryParams, extra: ExtraParams):
+    async def autocomplete(self, query: QueryParams, extra: ExtraParams):
         params = query.bragi_query_dict()
         body = None
         if extra.shape:
             body = extra.dict()
-        return self.raw_autocomplete(params, body)
+        response = await self.raw_autocomplete(params, body)
```
GuillaumeGomez: Isn't `return await ...` better?

amatissart: yes indeed

idunn/geocoder/nlu_client.py
```diff
@@ -119,44 +113,43 @@ def build_intention_category_place(self, tags_list, lang):
...
-                # 1 category
-                intention = self.build_intention_category(tags_list, lang=lang)
-                if intention is not None:
-                    intentions.append(intention)
-
-            return intentions
+
+        intentions = []
+        tags_list = response_nlu.json()["NLU"]
+        tags_list = [t for t in tags_list if t["tag"] != "O"]
```
GuillaumeGomez: Just realized: can't you do this directly when getting `tags_list` on the previous line?

amatissart: updated

tests/test_autocomplete.py
```diff
@@ -27,63 +27,40 @@ def read_fixture(sPath):
...
-        """
-        Two queries to the classifier API are required here,
-        as a consequence the order of the two queries below is important
-        """
-        mocked_responses.add(
-            responses.POST, f"{CLASSIF_URL}", json=FIXTURE_CLASSIF_pharmacy, status=200,
-        )
+        respx.post(NLU_URL, content=FIXTURE_TOKENIZER)
+        respx.post(CLASSIF_URL, content=FIXTURE_CLASSIF_pharmacy)
         yield mocked_responses
```
remi-dupre: As `mocked_responses` is not used anymore, isn't there a way to use respx's own decorator to avoid adding it?

PR #122, opened 2020-03-09
Handle all categories from NLU api

remi-dupre:
We were previously only checking if any of these two tags were applied to the user request: `cat` and `city`, and if at most 1 part of the request was applied to each of them.

The exhaustive list of tags returned from the NLU API is defined as follows:

```
- POI
- brand
- cat
- city
- country
- other
- state
- street
- O  (Unknown)
```

The proposed new behavior is:

 - if the label `poi` is part of the request, returns no intention (as the user is likely to search for a specific point)
 - if the label `other` is part of the request, we also return no intention to avoid unpredictable behavior
 - `brand` and `cat` are handled similarly (except that brands won't be input of the classifier), if there is several we also fallback to no intention
 - `city`, `country`, `state` and `street` are handled similarly, all parts of the requests labeled with them are concatenated

What would you think of this behavior?


remi-dupre:
We should no fallback to old behavior in the case were the tokenizer detects two brand and/or categories.

I'd still be interesting in writing a few more tests for this part of the code before merging this PR :)

idunn/geocoder/nlu_client.py
```diff
@@ -68,19 +73,25 @@ def fuzzy_match(cls, query, response):
...
-    async def build_intention_category_place(self, tags_list, lang):
-        cat_query = next(t["phrase"] for t in tags_list if t.get("tag") == "cat")
-        city_query = next(t["phrase"] for t in tags_list if t.get("tag") == "city")
+    async def build_intention_category_place(
+        self, cat_query, place_query, lang, skip_classifier=False
+    ):
+        async def get_category():
+            if skip_classifier:
+                return None
+            else:
```
GuillaumeGomez: No need for this else.

GuillaumeGomez: Euh actually, you never leave this part of the code... if and else parts both return.

idunn/geocoder/nlu_client.py
```diff
@@ -135,17 +178,31 @@ def fuzzy_match(cls, query, response):
...
+        if self.is_poi_request(tags_list):
+            return []
+
+        intentions = []
+        brand_query = self.build_brand_query(tags_list)
+        cat_query = self.build_category_query(tags_list)
+        place_query = self.build_place_query(tags_list)
+        skip_classifier = brand_query is not None
+
+        if (bool(brand_query) ^ bool(cat_query)) and place_query:
```
GuillaumeGomez: So if I understand correctly, you only enter in here if only one of brand_query or cat_query is true. I think it might be worth adding a comment to make the reading simpler.

remi-dupre: I just updated it with this comment:

```python
# If there is a label for both a category and a brand, we consider that the request is
# ambiguous and ignore both.
if bool(brand_query) ^ bool(cat_query):
    ...
```

idunn/geocoder/nlu_client.py
```diff
@@ -135,19 +177,36 @@ def fuzzy_match(cls, query, response):
...
+                intention = await self.build_intention_category_place(
+                    cat_query, place_query, lang=lang, skip_classifier=skip_classifier
+                )
+                if intention is not None:
+                    intentions.append(intention)
+            else:
+                # 1 category or brand
+                intention = await self.build_intention_category(
+                    cat_query, lang=lang, skip_classifier=skip_classifier
+                )
```
amatissart: `skip_classifier` is not a parameter for `build_intention_category`.  
It probably means one more test case should be added to cover that function.

amatissart: `brand_query` is never used to build the intention, instead `cat_query` is passed event if it's None.

remi-dupre: Oh, I missed this hack when witting tests: https://github.com/QwantResearch/idunn/blob/f5e00291a10c8bda992b7ea3f26069a2c98d2d39/tests/utils.py#L24.

I was afraid of adding more complicated code and it led me to forget about really testing brands :(

Is there a reason why we init PjSource globally? Maybe it could be more convenient to init it locally in modules where we use it? I suppose it would cost a few more http sessions which should be fine?

amatissart: We could probably initialize `PjSource` in different modules, but I am not sure it would make the tests any easier, given that settings would still be read on the first init only.

remi-dupre:
Most of the code should be covered by the tests now :)

idunn/api/geocoder.py
```diff
@@ -11,7 +11,12 @@
     async def get_intentions():
         if not query.nlu:
             return None
-        return await nlu_client.get_intentions(text=query.q, lang=query.lang)
+
+        focus = None
+        if query.lon and query.lat:
+            focus = (query.lat, query.lon)
```
amatissart: If I'm not wrong, these values are used as (x,y) to initialize a shapely `Point`, so they should be reversed.  
Maybe it would be clearer to create a `Point` instance here.

tests/test_autocomplete.py
```diff
@@ -142,6 +173,86 @@ def test_autocomplete_unavailable(mock_autocomplete_unavailable):
...
+        expected_intention_place=None,
+    )
+
+
+@enable_pj_source()
+def test_autocomplete_with_nlu_brand_focus(mock_autocomplete_get, mock_NLU_with_brand):
+    client = TestClient(app)
+    assert_ok_with(
+        client,
+        params={"q": "auchan", "lang": "fr", "limit": 7, "nlu": True, "lon": 48.9, "lat": 2.3},
```
amatissart: yes indeed, "lon" and "lat" are reversed ;)

idunn/api/geocoder.py
remi-dupre: Grrrrr, I still manage to get confused with this even when I try to look for it ... :angry:

remi-dupre: It should be fixed in 119d1a4  though

PR #125, opened 2020-03-18
Start implementation to get kuzzle info for trash

GuillaumeGomez:


idunn/places/poi.py
```diff
@@ -12,6 +13,19 @@ def __init__(self, d):
...
 
+        # If it's a trash can and in Bretagne (in France), we get its volume and the time of the
+        # last update
+        if (self.properties.get("amenity", "") == "recycling"
+            and any(x.get("label", "").lower().endswith("bretagne, france")
+                    for x in d.get("administrative_regions", []))):
+            try:
+                # TODO: use global timeout
+                res = requests.get(f"http://kuzzle:7512/<index>/<collection>/{self.id}", timeout=self.request_timeout)
+               self.properties["trash_volume"] = res.get("result", {}).get("_source", {}).get("volume")
```
amatissart: the indentation is broken here (see Travis reports)

idunn/places/poi.py
```diff
@@ -12,6 +13,19 @@ def __init__(self, d):
...
+               self.properties["trash_last_update"] = res.get("result", {}).get("_source", {}).get("hour")
```
amatissart: These new properties are specific to the `recycling` data, so it would probably be cleaner to build them in a new dedicated `RecyclingBlock` that will be necessary at some point to include this data in the Idunn response.

idunn/places/poi.py
```diff
@@ -12,6 +13,19 @@ def __init__(self, d):
             self["properties"] = {p.get("key"): p.get("value") for p in self.get("properties", [])}
         self.properties = self["properties"]
 
+        # If it's a trash can and in Bretagne (in France), we get its volume and the time of the
+        # last update
+        if (self.properties.get("amenity", "") == "recycling"
+            and any(x.get("label", "").lower().endswith("bretagne, france")
+                    for x in d.get("administrative_regions", []))):
```
amatissart: It should be possible to use the different `codes` present in each administrative region, instead of relying on string comparison.  

For example for Finistère:
```
"codes": [
  {"name":"ISO3166-2","value":"FR-29"},
  {"name":"ref:INSEE","value":"29"},
  {"name":"ref:NUTS","value":"FR522"},
  {"name":"wikidata","value":"Q3389"}
]
```
The `wikidata` entity or the `ISO3166-2` code could be used as reference.

GuillaumeGomez: I found this way of doing it simpler and easier to read.

amatissart:
Included in #136

PR #126, opened 2020-03-26
add covid19 block

GuillaumeGomez:


idunn/blocks/covid19.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @classmethod
+    def from_es(cls, es_poi, lang):
+        if settings["BLOCK_COVID_ENABLED"] is not True:
+            return None
+        opening_hours = es_poi.properties.get("opening_hours:covid19")
+        status = "unknown"
+        if opening_hours == "same":
+            opening_hours = OpeningHourBlock.from_es(es_poi.properties.get("opening_hours"), lang)
+            status = "open_as_usual"
```
amatissart: `open_as_usual` should be used too if both tags `opening_hours:covid19` and `opening_hours` are equal

idunn/blocks/covid19.py
```diff
@@ -0,0 +1,40 @@
...
+            opening_hours = None
+            status = "closed"
+        elif opening_hours is not None:
+            opening_hours = OpeningHourBlock.from_es(opening_hours, lang)
+            if opening_hours is None:
+                status = "unknown"
+            elif opening_hours.status == "open":
+                status = "open"
+            elif opening_hours.status == "closed":
+                status = "closed"
```
amatissart: Here `opening_hours.status` represents the realtime status of the POI.
I think the covid19 block `status` should be `open` in both cases, meaning that the POI is not completely closed during the lockdown.

GuillaumeGomez: Then what am I supposed to put in here?

PR #127, opened 2020-03-27
Covid19 block: load additional data from caresteouvert dataset

amatissart:
* Use [a background task](https://fastapi.tiangolo.com/tutorial/background-tasks/) to fetch the dataset regularly and store data in Redis
* This dataset is used as a fallback if no status has been found thanks to OSM tags
* `contribute_url` contais a direct URL for the current POI on caresteouvert.fr


idunn/blocks/covid19.py
```diff
@@ -1,45 +1,63 @@
...
     STATUSES: ClassVar = ["open_as_usual", "open", "maybe_open", "closed", "unknown"]
 
-    status: constr(regex="({})".format("|".join(STATUSES)))
+    status: constr(regex="^({})$".format("|".join(STATUSES)))
     opening_hours: Optional[OpeningHourBlock]
     note: Optional[str]
+    contribute_url: Optional[str]
+
+    @classmethod
+    def get_caresteouvert_link(cls, place):
```
GuillaumeGomez: For readability reason, I think you should rename it "get_ca_reste_ouvert_link".

amatissart: done

PR #129, opened 2020-04-09
Fix: images block fails when POI has no name

amatissart:
The `alt: str` property for images is derived from the poi name.
But the case when no name is available was not handled.

### DIFF
--- a/tests/test_wiki_images.py
+++ b/tests/test_wiki_images.py
@@ -2,35 +2,9 @@ import os
 import json
 import pytest
 from app import app
+from unittest import mock
+from idunn.places import POI
 from starlette.testclient import TestClient
-
-INDICES = {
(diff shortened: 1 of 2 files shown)


tests/test_wiki_images.py
```diff
@@ -2,36 +2,10 @@
...
-            doc_type=doc_type,  # 'admin', 'street', 'addr' or 'poi'
-            id=place_id,
-            refresh=True,
-        )
-
-
-@pytest.fixture(autouse=True)
-def load_all(mimir_client, init_indices):
-    load_place("orsay_museum.json", mimir_client, "poi")
-
```
amatissart, line 34: This code was unused as the fixtures are already defined globally in conftest.py

PR #130, opened 2020-04-20
Place model: add 'country_code' in address object

amatissart:
* Define `Address` model, used in existing `Place` response model
* Implement `.get_country_code()` for all places  
This method returns the least specific country code among all codes defined in administrative regions.
* Add `country_code` field in address objects

### DIFF
--- a/idunn/places/event.py
+++ b/idunn/places/event.py
@@ -32,7 +32,6 @@ class Event(BasePlace):
         return {
             "name": self.get("placename"),
             "label": self.get("address"),
-            "city": self.get("city"),
             "admin": self.build_admin(lang),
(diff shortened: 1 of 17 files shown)


idunn/places/event.py
```diff
@@ -32,7 +32,6 @@ def build_address(self, lang):
         return {
             "name": self.get("placename"),
             "label": self.get("address"),
-            "city": self.get("city"),
```
amatissart, line 35: This field `city` was specific to `Event` objects, and does not fit the `Address` model used by other objects.  
It's not used by any application, and the city name is already part of the label. 
So there is no reason to keep it.

PR #131, opened 2020-04-22
Consolidate POI status related to covid-19, improve error handling

amatissart:
_These 3 commits may be reviewed separately_

* More explicit messages about Redis errors 
* Clean opening status logic in covid19 block
  * use Redis data in priority (maintained by CaResteOuvert, and refreshed more regularly)
  * handle "open" and "restricted" values for `opening_hours:covid19`
* Define a list of country codes in settings where covid19 block should be enabled

### DIFF
--- a/idunn/utils/redis.py
+++ b/idunn/utils/redis.py
@@ -46,8 +46,7 @@ class RedisWrapper:
             prometheus.exception("RedisError")
             if raise_on_error:
                 raise
-            else:
-                logging.exception("Got a RedisError")
+            logging.exception("Got a RedisError")
 
     @classmethod
     def _get_value(cls, key):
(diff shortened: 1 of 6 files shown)


idunn/utils/redis.py
```diff
@@ -39,12 +39,15 @@ class RedisWrapper:
...
-    def _set_value(cls, key, value, expire=settings["WIKI_CACHE_TIMEOUT"]):
+    def _set_value(cls, key, value, expire=settings["WIKI_CACHE_TIMEOUT"], raise_on_error=False):
         try:
             cls._connection.set(key, value, ex=expire)
         except RedisError:
             prometheus.exception("RedisError")
-            logging.exception("Got a RedisError")
+            if raise_on_error:
+                raise
+            else:
```
GuillaumeGomez, line 49: I guess the `else` is unneeded here?

amatissart: indeed. It's corrected.

PR #135, opened 2020-05-12
NLU: allow full-text query outside France and without focus

amatissart:
In the first implementation, an `Intention` with full-text query was returned only when `pj_source` was available for the current focus point (i.e in France).

Now that #134 introduces full-text queries for all data sources, this condition can be removed. 

As a consequence, a query with a brand name may also be returned if no focus is set. We need to check if this behavior is expected and that the front-end application will behave correctly in this case.

### DIFF
--- a/tests/test_autocomplete.py
+++ b/tests/test_autocomplete.py
@@ -186,7 +186,7 @@ def test_autocomplete_with_nlu_brand_no_focus(mock_autocomplete_get, mock_NLU_wi
     assert_ok_with(
         client,
         params={"q": "auchan", "lang": "fr", "limit": 7, "nlu": True},
-        expected_intention=[{"filter": {"q": "auchan"}, "description": {"query": "auchan"},}],
+        expected_intention=[{"filter": {"q": "auchan"}, "description": {"query": "auchan"}}],
         expected_intention_place=None,
     )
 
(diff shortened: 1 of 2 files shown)


remi-dupre:
We could probably also allow to return a "full text query + place outside of france" [here](https://github.com/QwantResearch/idunn/pull/135/files#diff-1850bba4a90dd4cc280007845eef9acbL119-L124)?
```python
        if category_name:
            return Intention(
                filter={"category": category_name, "bbox": bbox},
                description={"category": category_name, "place": place},
            )
        else:
            if not pj_source.bbox_is_covered(bbox):
                return None
            return Intention(
                filter={"q": cat_query, "bbox": bbox},
                description={"query": cat_query, "place": place},
            )

**Comment by amatissart** [comment]
> We could probably also allow to return a "full text query + place outside of france" [here](https://github.com/QwantResearch/idunn/pull/135/files#diff-1850bba4a90dd4cc280007845eef9acbL119-L124)?

Good catch, the PR has been updated.

File: tests/test_autocomplete.py [hunk 1]
```diff
@@ -186,7 +186,7 @@ def test_autocomplete_with_nlu_brand_no_focus(mock_autocomplete_get, mock_NLU_wi
     assert_ok_with(
         client,
         params={"q": "auchan", "lang": "fr", "limit": 7, "nlu": True},
-        expected_intention=[],
+        expected_intention=[{"filter": {"q": "auchan"}, "description": {"query": "auchan"},}],
```
remi-dupre, line 189: Nobody cares but ...
```suggestion
        expected_intention=[{"filter": {"q": "auchan"}, "description": {"query": "auchan"}}],
```
Y a une virugule en trop ... là ... :'D
(This is due to black's behaviour)

amatissart: fixed :D

PR #136, opened 2020-05-18
Recycling block with multiple containers and test data

amatissart:
Rebased and extended version of #125 with details about multiple waste containers.

On this branch, the recycling block returns fake data for testing purposes: 3 different containers for all "recycling" POIs in Finistère.

### DIFF
--- a/idunn/blocks/recycling.py
+++ b/idunn/blocks/recycling.py
@@ -1,24 +1,87 @@
-from .base import BaseBlock
-
-from typing import ClassVar
+from datetime import datetime, timedelta
+from typing import ClassVar, List
+from enum import Enum
+import requests
+from pydantic import BaseModel, validator
+from pytz import UTC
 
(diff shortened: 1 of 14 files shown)


idunn/blocks/recycling.py
```diff
@@ -0,0 +1,100 @@
...
+            # Data source is not configured
+            return None
+
+        if place.PLACE_TYPE != "poi":
+            return None
+
+        if place.get_class_name() != "recycling":
+            return None
+
+        if not is_poi_in_finistere(place):
```
GuillaumeGomez, line 55: Can't you make the three conditions into one?

amatissart: These conditions will change very soon, and should probably be set in the configuration (for example to allow multiple zones where this block should be enabled).

I will keep them for now, waiting for more specifications.

PR #146, opened 2020-06-11
Implement "extend_bbox" parameter on /places endpoint

amatissart:
## Description

* `/v1/places` endpoint accepts a new boolean parameter: `extend_bbox` (false by default). If this parameter is set to true and no results are found in the original `bbox`, a second query will be executed on an extended bbox and its results will be returned instead.
* In addition to the `places` list and `source`, 2 new fields are present in the response:
  * `bbox`: The minimal bbox containing all results. `null` if no place is found.  
It may be **larger** than or **outside** of the original bbox passed in the query if `extend_bbox=true` was set.
  * `bbox_extended` (bool): true if `extend_bbox=true` was set and a second search has been executed, after no results were found in the original bbox included in the URL.

## Why
This new option should enable the front-end application to adapt the view to show relevant results on the map, without manual "zoom out" by the user.

### DIFF
--- a/idunn/api/places_list.py
+++ b/idunn/api/places_list.py
@@ -1,7 +1,8 @@
 import os
 import logging
 
-from shapely.geometry import MultiPoint
+from shapely.geometry import MultiPoint, box
+from shapely.affinity import scale
 from fastapi import HTTPException, Query
 from fastapi.concurrency import run_in_threadpool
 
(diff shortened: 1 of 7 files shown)


idunn/api/places_list.py
```diff
@@ -152,86 +169,109 @@ def valid_categories(cls, v):
...
+    bbox_extended = False
+
+    if params.extend_bbox and len(places_list) == 0:
+        original_bbox = params.bbox
+        original_bbox_width = original_bbox[2] - original_bbox[0]
+        original_bbox_height = original_bbox[3] - original_bbox[1]
+        original_bbox_size = max(original_bbox_height, original_bbox_width)
+        if original_bbox_size < EXTENDED_BBOX_MAX_SIZE:
+            # Compute extended bbox and fetch results a second time
+            scale_factor = EXTENDED_BBOX_MAX_SIZE / original_bbox_size
```
bbecquet, line 226: Don't we have a spatial library to do this kind of error-prone operations for us?

amatissart: You're right. Using [`shapely.affinity.scale`](https://shapely.readthedocs.io/en/latest/manual.html#shapely.affinity.scale) makes things a little simpler.

PR #147, opened 2020-06-16
add more logs for intention results

remi-dupre:
Add one line of logs each time intention is checked, which should be convenient to build metrics about it usage.

I integrated as much details as possible about results of external calls in case we want to performed more refined analysis. I don't think I am missing any exciting data to add?

### DIFF
--- a/idunn/geocoder/nlu_client.py
+++ b/idunn/geocoder/nlu_client.py
@@ -194,23 +194,23 @@ class NLU_Helper:
         return response_nlu
 
     async def get_intentions(self, text, lang, focus=None):
+        logs_extra = {
+            "intention_detection": {
+                "text": text,
+                "lang": lang,
+                "focus": mapping(focus).get("coordinates") if focus is not None else None,
+            }
+        }
+
         try:
             response_nlu = await tagger_circuit_breaker.call_async(
                 self.post_intentions, text, lang, focus
             )
         except Exception:
-            logger.error("Request to NLU tagger failed", exc_info=True)
+            logger.error("Request to NLU tagger failed", exc_info=True, extra=logs_extra)
             return []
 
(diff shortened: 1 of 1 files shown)


idunn/geocoder/nlu_client.py
```diff
@@ -204,6 +204,11 @@ async def get_intentions(self, text, lang, focus=None):
         tags_list = [t for t in response_nlu.json()["NLU"] if t["tag"] != "O"]
 
         if self.is_poi_request(tags_list):
+            logger.info(
+                "Detected POI request for '%s'",
+                text,
+                extra={"lang": lang, "focus": focus, "tags_list": tags_list},
```
amatissart, line 210: `text` would be useful in the `extra` object too, to make the extraction easier.

Also, all keys in `extra` appear in the log message at the root level (when using json format for logs).
I think it would be more appropriate to nest this data under a dedicated key, such as `intention_detection` ?

`focus` is a shapely `Point`, that should be serialized in an explicit way.

idunn/geocoder/nlu_client.py
```diff
@@ -237,6 +242,22 @@ async def get_intentions(self, text, lang, focus=None):
...
+        logs_extra = {
+            "lang": lang,
+            "focus": focus,
+            "tags_list": tags_list,
+            "intentions": [x.dict() for x in intentions],
+            "brand_query": brand_query,
+            "cat_query": cat_query,
+            "place_query": place_query,
+            "skip_classifier": skip_classifier,
+        }
```
amatissart, line 254: Same as above.

Moreover, I am not sure about using the raw `intentions` list in the log message. It could make the mapping of the data storage for these logs hard to maintain.

remi-dupre: I just removed it then, other fields are probably light enough (and generally most of them are empty)

PR #162, opened 2020-08-18
add radius parameter for the geocoder

remi-dupre:
Take benefits of https://github.com/CanalTP/mimirsbrunn/pull/417 to allow a larger focus mode with larger zoom levels.

The endpoint `autocomplete` takes a new optional parameter `zoom` to pass the current zoom level of the front-end, which is used to decide the typical expected distance of results. I am using [this branch](https://github.com/QwantResearch/erdapfel/pull/743) of erdapfel to test this change:

| Z = 3 | Z = 5 | Z = 7 | Z = 9 | Z = 11 |
|:----------------:|:-----:|-------|-------|-------------------------------|
| Screenshot_2020-08-19_15-49-50 | Screenshot_2020-08-19_15-50-11 | Screenshot_2020-08-19_15-50-38 | Screenshot_2020-08-19_15-50-58 | Screenshot_2020-08-19_15-51-17 | 

### Maybe some of the logic should be moved in erdapfel?

The correspondence between zoom level and radius parameter of the geocoder is defined in this PR, but one could argue that it should be implemented in the front end:

 - we could decide to set a shorter radius at equivalent zoom level on smaller devices
 - this doesn't really play well with [erdapfel setting its minimum zoom value for the focus](https://github.com/QwantResearch/erdapfel/blob/master/src/libs/suggest.js#L16), this is probably a behavior we want to keep in order not to affect the cache efficiency.

### DIFF
--- a/idunn/geocoder/models/params.py
+++ b/idunn/geocoder/models/params.py
@@ -2,6 +2,7 @@
 Bragi parameters as defined here:
   https://github.com/CanalTP/mimirsbrunn/blob/master/libs/bragi/src/routes/autocomplete.rs#L59
 """
+import json
 from enum import Enum
 from typing import List, Optional
 from fastapi import Query
@@ -10,6 +11,9 @@ from pydantic.dataclasses import dataclass
 
 from idunn import settings
 from .cosmogony import ZoneType
+
+focus_zoom_to_radius = json.loads(settings["FOCUS_ZOOM_TO_RADIUS"])
+minimum_zoom_for_focus = min(zoom for zoom, _ in focus_zoom_to_radius)
 
 
 class Type(str, Enum):
@@ -47,8 +51,6 @@ class QueryParams:
         """
         params = {
             "q": self.q,
-            "lon": self.lon,
-            "lat": self.lat,
             "lang": self.lang,
             "limit": self.limit,
             "pt_dataset[]": self.pt_dataset,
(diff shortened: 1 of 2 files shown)


bbecquet:
> * we could decide to set a shorter radius at equivalent zoom level on smaller devices
>

Interesting point… Sending the device size to the server would feel probably too much like device fingerprinting, but we could send the visible bbox as another parameter?
It seems there are many approches to this problem, but as this PR adds more intelligence on the server, this is probably where we should continue building stuff instead of splitting the logic between both sides.

> * this doesn't really play well with erdapfel setting its minimum zoom value for the focus, this is probably a behavior we want to keep in order not to affect the cache efficiency.
> 

I don't know enough the back-end constraints, but I'm not sure cache was the main reason for it.
What level of cache are we talking about anyway?

remi-dupre:
> Interesting point… Sending the device size to the server would feel probably too much like device fingerprinting, but we could send the visible bbox as another parameter?

That should probably work, I don't think that it would be a critical improvement that we want to implement right now anyway.

It is also interesting that a bbox is more concrete concept than an arbitrary definition of zoom.

> It seems there are many approches to this problem, but as this PR adds more intelligence on the server, this is probably where we should continue building stuff instead of splitting the logic between both sides.

Yeah that's one of the reasons why I naturally implemented it here without thinking about it.

> I don't know enough the back-end constraints, but I'm not sure cache was the main reason for it.
> What level of cache are we talking about anyway?

Actually, this setting previously existed because the front-end was in charge of enabling the focus (by sending lat/lon). We could just remove this setting in order to always send lat/lon/zoom to Idunn. If the cache is of any concern we should make sure that we send less precise coordinates with higher levels of zoom, and we would probably be fine.

bbecquet (approved):

amatissart (commented):
About the minimal zoom value where the focus mode is used:
 * it is actually defined as a constant (11) in Erdapfel
 * the main reason for it was to keep deterministic ranking for most requests, as most of our tests suites do not use the focus mode and using the focus mode gives less weight to individual item importance (admins population and POIs popularity). The new behavior where the focus mode would be used for most requests remains to be tested more exhaustively.
 * with the implementation of this PR, I agree we could indeed remove this zoom condition in Erdapfel and let Idunn be responsible for activating the focus mode if it's relevant. It would not directly affect the cache behavior, as only requests to Bragi (now called by Idunn) are cached. So Idunn could be in charge of rounding the coordinates and build deterministic query parameters to optimize the cache behavior. However we should still be cautious about the precision of the data used in the request to Idunn (to avoid sending unnecessary data to the server about the user activity).

remi-dupre:
Should be fine now :)

amatissart (approved):

PR #166, opened 2020-09-18
nlu: Add new categories, and allow some categories to be classified from 'brand' queries

amatissart:
* The NLU client skips the call to the classifier for "brand" queries. However we noticed the tagger may not make the distinction between "category" and "brand" reliably.  
As a workaround, this PR suggests to remove the `skip_classifier` mechanism, and to always call the "regex" classifier, and try to classify the brand query with a subset of the list of categories. These categories are tagged with `match_brand: True` in "categories.yml".

* `recycling` is currently the only category with `match_brand: True`

* Add 3 new categories: `station`, `parking`, and `place_of_worship`.

### DIFF
--- a/idunn/geocoder/nlu_client.py
+++ b/idunn/geocoder/nlu_client.py
@@ -61,18 +61,20 @@ class NLU_Helper:
 
         return response_classifier.json()["intention"][0][1]
 
-    def regex_classifier(self, text):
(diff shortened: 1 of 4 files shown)


remi-dupre (approved):
This looks like a very relevant change until the tagger becomes good enough to rely exclusively on it :)

PR #171, opened 2020-10-16
opening hours: Ignore public holidays in countries where opening_hours.js does not support them

amatissart:
"opening_hours.js" support public holidays in a certain list of countries: https://github.com/opening-hours/opening_hours.js#holidays

Opening hours evaluation should ignore PH clauses in the other countries, to return opening hours for a typical week.

### DIFF
--- a/idunn/utils/opening_hours.py
+++ b/idunn/utils/opening_hours.py
@@ -1,7 +1,7 @@
 import os
 from datetime import date, datetime, time, timedelta
 
-from py_mini_racer.py_mini_racer import MiniRacer
+from py_mini_racer.py_mini_racer import MiniRacer, JSEvalException
 
 DIR = os.path.dirname(__file__)
 OPENING_HOURS_JS = os.path.join(DIR, "data/opening_hours.min.js")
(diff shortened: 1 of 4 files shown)


idunn/utils/opening_hours.py
```diff
@@ -37,16 +37,20 @@ def call(self, *args, **kwargs):
...
+        else:
+            self.nmt_obj = None
 
     def validate(self):
         """Check if an expression parses correctly"""
-        return engine.call("validate", self.raw, self.nmt_obj) is True
+        result = engine.call("validate", self.raw, self.nmt_obj)
+        if self.nmt_obj is not None and result is not True:
+            self.nmt_obj = None
+            return self.validate()
```
remi-dupre, line 52: I feels weird to hide any exception we could get when `nmt_obj` is not `None`, maybe it is possible to explicitly specify the kind of error we expect here?

amatissart: That's a good point.  

But openinghours.js throws raw error strings such as 
```
An error occurred during evaluation of the value "We-Su 17:00-21:00; PH off". Please submit a pull request: https://github.com/opening-hours/opening_hours.js. There are no holidays (PH) defined for country jp.
```
So I suppose we could catch the error and revalidate only if `"no holidays"` is contained in the string :worried: 
Any better idea ?

idunn/utils/opening_hours.py
```diff
@@ -46,11 +46,16 @@ def __init__(self, oh, tz, country_code):
...
     def validate(self):
         """Check if an expression parses correctly"""
-        result = engine.call("validate", self.raw, self.nmt_obj)
-        if self.nmt_obj is not None and result is not True:
-            self.nmt_obj = None
-            return self.validate()
-        return result is True
+        try:
+            engine.call("validate", self.raw, self.nmt_obj)
+        except JSEvalException as exc:
```
remi-dupre, line 51: Very cool that exceptions are translated by pyminiracer!

remi-dupre (approved):

PR #172, opened 2020-10-30
Exclude implausible intentions with full-text 'q' only and few matching features

amatissart:
This is an attempt to reduce false positives caused by queries wrongly tagged as "brand" by the tagger.

By comparing the query with the features `name`, it's possible to evaluate if the current query does effectively match multiple features among the results provided by the geocoder. A real brand name will typically match dozens of POIs.

This filter is not applied if the intention contains both a query and a geocoded place. False positives are less likely in this case.


remi-dupre (approved):
That sounds like a nice idea.

Do you think we should keep hope that will remove these kind of feature some day and ease it with some kind of feature flag?

amatissart:
A feature flag is now defined. And a missing fixture has been committed.

idunn/api/geocoder.py
```diff
@@ -51,10 +52,12 @@ async def get_intentions():
...
     if intentions is not None and query.nlu:
-        intentions = list(
-            filter(
-                lambda i: post_filter_intention(i, autocomplete_response, query.limit), intentions
+        if autocomplete_nlu_filter_intentions:
+            intentions = list(
+                filter(
+                    lambda i: post_filter_intention(i, autocomplete_response, query.limit),
+                    intentions,
+                )
```
remi-dupre, line 60: I'm never sure what's more readable between list comprehension and `list(filter(...))` + lambda. Python has just too many reasonable alternatives :p

```suggestion
        intentions = [
            i
            for i in intentions
            if post_filter_intention(i, autocomplete_response, query.limit)
        ]
```

Not very important, just something I've had on my mind recently...

remi-dupre (approved):
Sounds neat!

idunn/api/geocoder.py
amatissart: I should definitely use list comprehension more systematically. Less natural to write in my opinion, but more readable probably.

PR #174, opened 2020-11-25
API Pages Jaunes

remi-dupre:
Some notes about the PR itself:

 - [x] PJ doesn't provide a specific field for the house number, I don't think we use it anyway?
 - [x] There is a separate field for thumbnails, it could be nice to handle it separately (show it in list results but not in description images)
 - [ ] I handled wheelchair accessibility with what seems to be a fair approximation ([sometimes better](https://www.qwant.com/maps/place/pj:07383963#map=16.50/48.8452000/2.3741831)), but there is still some POIs for which I don't understand where the information could come from. For example [57192322](https://www.qwant.com/maps/place/pj:57192322#map=16.50/48.1009870/-2.0312209) looks like [this](https://gist.github.com/remi-dupre/a8208ff96aa34a7c5bea4c5b3a367a20).
 - [ ] PJ can only return 30 results per query, as the front-end query for 40 by default we need to make two requests, it could be nice to make them asynchronously (but it think we should leave it for a future PR)

Some issues I saw while using the API:

 - [ ] The "Business Infos" API doesn't give the geographic location of the API 
 - [ ] There is many tiny inconsistencies between the two APIs (eg. `contact_info` in find API = `contact_infos` in infos API) which could make it consequently easier to use if fixed
 - [ ] `next_page_url` redirects to https://api.pagesjaunes.fr/v1/pros/find instead of https://api.pagesjaunes.fr/v1/pros/search
 - [x] The website URL is routed through PJ which explicitly intends to collect stats
 - ~~There is no refresh token yet where it seems that it is intended [in the doc](https://developer.pagesjaunes.fr/quickstart)~~

Some of these questions may require a direct discussion with PJ, I'm just putting it here until the implementation is closer to its final state.


idunn/places/pj_poi.py
```diff
@@ -210,3 +215,185 @@ def get_raw_grades(self):
...
+class PjApiPOI(BasePlace):
+    PLACE_TYPE = "poi"
+
+    def __init__(self, d: Union[pj_find.Listing, pj_info.Response]):
+        super().__init__(d)
+        self.data = d
+
+    def get_id(self):
+        listing_id = self.data.listing_id
+        return f"pj:{listing_id}" if listing_id else None
```
amatissart, line 229: The small id used by Qwant Maps is actually the `merchant_id` (composed of 8 digits).

idunn/datasources/pages_jaunes.py
```diff
@@ -58,21 +77,78 @@ def get_places_bbox(self, raw_categories, bbox, size=10, query=""):
...
+        if not res:
+            return None
+
+        return [PjApiPOI(listing) for listing in res.search_results.listings or []]
+
+    def get_place(self, poi_id) -> PjApiPOI:
+        return PjApiPOI(
+            pj_info.Response(
+                **self.get_from_params(
+                    self.PJ_INFO_API_URL, {"listing_id": self.internal_id(poi_id)}
```
amatissart, line 148: Same comment: `poi_id` is based on the `merchant_id`, that needs to be passed in the URL path (and not as a query param).

remi-dupre: Oh right, then the business info API has to be called using the positional `pro_id` parameter (which appears to be equivalent to `merchant_id`? I'll add a comment in my list), that's what go me confused as `?merchant_id=XXXXXXXX` leads to empty results.

idunn/datasources/pages_jaunes.py
```diff
@@ -58,21 +77,78 @@ def get_places_bbox(self, raw_categories, bbox, size=10, query=""):
...
+    def get_places_bbox(self, raw_categories, bbox, size=10, query="") -> List[PjApiPOI]:
+        query_params = {
+            "what": raw_categories,
+            "where": self.format_where(bbox),
+            # TODO: add some scrolling mechanism, which should come with some async?
+            "max": min(30, size),
+        }
+
+        if query:
+            query_params["q"] = query
```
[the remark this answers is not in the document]
remi-dupre: Nice catch! Our use of `query` is actually exactly matching the use case of `what`, right?

idunn/datasources/pages_jaunes.py
```diff
@@ -58,21 +77,78 @@ def get_places_bbox(self, raw_categories, bbox, size=10, query=""):
...
+
+        # TODO: refresh token if available?
+        #       https://requests-oauthlib.readthedocs.io/en/latest/oauth2_workflow.html#refreshing-tokens
+        self.client = BackendApplicationClient(client_id=self.client_id)
+        self.oauth = OAuth2Session(client=self.client)
+        self.token = self.oauth.fetch_token(
+            token_url=self.PJ_AUTHORIZATION,
+            client_id=self.client_id,
+            client_secret=self.client_secret,
+        )
```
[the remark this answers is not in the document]
remi-dupre: I'll have a look into it.

I am not sure PJ doesn't intend to provide a refresh token in the future as several fields related to it are provided (`refresh_token_expires_in` and `refresh_count`) and mostly because the [quickstart](https://developer.pagesjaunes.fr/quickstart) page of the documentation has a section "Refresh access token" in the table of contents which doesn't seem to be filled yet. I guess requesting a new token by hand is very simple and clean enough even if an alternative mechanism would be provided?

idunn/api/utils.py
```diff
@@ -363,7 +363,8 @@ def get_name(properties, lang):
 
 def parse_phone_number(phone):
     try:
-        return phonenumbers.parse(phone)
+        # TODO: mechanism for region
+        return phonenumbers.parse(phone, region="FR")
```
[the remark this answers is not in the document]
remi-dupre: PJ provides phone numbers without the country prefix, a cheap workaround would be to pop a "0" and add the expected "+33",  do you think it's better for now?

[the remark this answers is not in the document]
amatissart: Solved by #175

idunn/datasources/pages_jaunes.py
```diff
@@ -58,21 +75,87 @@ def get_places_bbox(self, raw_categories, bbox, size=10, query=""):
...
+                size=size - len(pois),
+            )
+
+        return pois
+
+    def get_places_bbox(self, raw_categories, bbox, size=10, query="") -> List[PjApiPOI]:
+        query_params = {
+            "what": " ".join(raw_categories),
+            "where": self.format_where(bbox),
+            "max": min(self.PJ_RESULT_MAX_SIZE, size),
```
amatissart, line 145: Note: before merging we need to decide whether we should keep this check, or whether successive calls to fetch several pages are acceptable

idunn/datasources/recycling.py
```diff
@@ -5,23 +5,47 @@
...
+        }
+
+    @staticmethod
+    def parse_authorisation_response(resp):
+        token = resp["result"]["jwt"]
+        expires_at = resp["result"]["expiresAt"] / 1000
+        return token, expires_at
+
+    def query_new_token(self):
+        return self.inner.post(
```
amatissart, line 32: the authentication URL expects a `GET` request

idunn/places/models/pj_find.py
```diff
@@ -0,0 +1,329 @@
...
+
+
+class Type(Enum):
+    RESERVER = "RESERVER"
+    RESERVER_INTERNE = "RESERVER_INTERNE"
+    COMMANDER = "COMMANDER"
+    COMMANDER_CHRONO = "COMMANDER_CHRONO"
+    PRENDRE_RDV_EXTERNE = "PRENDRE_RDV_EXTERNE"
+    PRENDRE_RDV_INTERNE = "PRENDRE_RDV_INTERNE"
+
```
amatissart, line 155: Side note: should we remove/comment these unused models ?  
Any breaking change in the API could break our app with no reason.

tests/test_pj_poi.py
```diff
@@ -1,84 +1,113 @@
...
+                places_utils.pj_source, "get_from_params", new=lambda *x, **y: api_result,
+            )
+
+        with api_mock:
+            yield
+
+
+@pytest.mark.parametrize(
+    "enable_pj_source", [("legacy", "musee_picasso"), ("api", "api_musee_picasso")], indirect=True,
+)
```
amatissart, line 48: cool

idunn/utils/auth_session.py
```diff
@@ -0,0 +1,71 @@
...
+class AuthSession:
+    """
+    Helper class for HTTP sessions that need to keep an authentification token.
+    Default behavior corresponds to an OAuth2 API.
+
+    Note that at least `get_authorization_url` and `get_authorization_params`
+    need to be overriden.
+    """
+
+    def __init__(self, expiration_tolerance=10, refresh_timeout=1000):
```
amatissart, line 16: A bit confusing to have 2 parameters with different units (s, ms). At least a docstring would be welcome.

idunn/datasources/pages_jaunes.py
remi-dupre: This max is the query parameter sent to the API, although current implementation will apply the successive calls in `get_places_from_url` by opening recursively `next_page_url`

idunn/utils/auth_session.py
remi-dupre: Oh crap, that's indeed a really bad idea, I must have put it on hold and forgotten about it later :/

idunn/places/models/pj_find.py
remi-dupre: At least it would be easy to update it by just re-downloading their openapi model :/

I'm not particularly convinced by both approach, neither "just not using any model at all". I'll remove unused parts as it feels indeed more robust.

remi-dupre: I removed unnecessary parts. This means that we are not afraid to stick to generated model so I took benefit of it by replacing all `Optional[List[...]]` by list fields which default to `[]` as this is slightly easier to work with :+1:

tests/test_pj_poi.py
```diff
@@ -1,84 +1,115 @@
...
+    resp = response.json()
+    assert resp["id"] == "pj:05360257"
+    assert resp["name"] == "Musée Picasso"
+    assert resp["address"]["label"] == "5 r Thorigny, 75003 Paris"
+    assert resp["class_name"] == "museum"
+    assert resp["subclass_name"] == "museum"
+    assert resp["type"] == "poi"
+    assert resp["meta"]["source"] == "pages_jaunes"
+    # TODO
+    # assert resp["geometry"]["center"] == [2.362634, 48.859702]
```
amatissart, line 63: The test could pass now, once the fixture has been updated

tests/test_pj_poi.py
```diff
@@ -1,84 +1,115 @@
...
+    resp = response.json()
+    assert resp["id"] == "pj:05360257"
+    assert resp["name"] == "Musée Picasso"
+    assert resp["address"]["label"] == ""
+    assert resp["class_name"] == "museum"
+    assert resp["subclass_name"] == "museum"
+    assert resp["type"] == "poi"
+    assert resp["meta"]["source"] == "pages_jaunes"
+    # TODO
+    #  assert resp["geometry"]["center"] == [2.362634, 48.859702]
```
amatissart, line 115: same

idunn/datasources/pages_jaunes.py
```diff
@@ -58,21 +75,87 @@ def get_places_bbox(self, raw_categories, bbox, size=10, query=""):
...
+            pois += self.get_places_from_url(
+                res.context.pages.next_page_url.replace("/pros/find", "/pros/search"),
+                size=size - len(pois),
+            )
+
+        return pois
+
+    def get_places_bbox(self, raw_categories, bbox, size=10, query="") -> List[PjApiPOI]:
+        query_params = {
+            "what": " ".join(raw_categories),
```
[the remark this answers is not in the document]
remi-dupre: This sounds like a reasonable approach, there is indeed no reason to keep this list for now.

To be sure I understand what you mean, in this case we would do something like that ?

```yaml
 education: 
     pj_filters: "école collège lycée enseignement supérieur"
```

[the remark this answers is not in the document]
amatissart: yes, but under a new name to preserve the old usage of `pj_filters` in ElasticSearch queries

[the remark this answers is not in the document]
remi-dupre: What would you think of `pj_what` to stay close to the API?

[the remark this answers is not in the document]
amatissart: nice :+1:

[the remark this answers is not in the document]
amatissart: As discussed, this specific issue with category filters can be solved separately.

amatissart (approved):

PR #175, opened 2020-11-30
Add more context about phone numbers parsing errors

amatissart:
### To Do
- [x] Pass the country code to the parsing function, to allow non international formats (Thanks @remi-dupre )

### DIFF
--- a/idunn/api/utils.py
+++ b/idunn/api/utils.py
@@ -27,7 +27,6 @@ from idunn.blocks import (
 from idunn.utils import prometheus
 from idunn.utils.index_names import INDICES
 from idunn.utils.es_wrapper import get_elasticsearch
-import phonenumbers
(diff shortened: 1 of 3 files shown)


remi-dupre:
I just pushed a [commit to handle country code](https://github.com/QwantResearch/idunn/pull/175/commits/22c09e6e3a5c2e57bc43450de9dbb38f8a48bfb9), I also removed the `get_formatted_phone_number` function as it seems that `phonenumbers.format_number` cannot fail (am I right? we should probably check the logs)

amatissart:
I just checked the logs, and I could not find any error on `get_formatted_phone_number`. And it looks safe according to its doc.  
So everything looks good to me :+1:

remi-dupre (approved):

PR #177, opened 2020-12-07
PJ: urlsover

remi-dupre:
Add a `.../v1/url` endpoint which can be passed an URL supposed to redirect to another URL. This will open the given URL from Idunn and return a redirection response to mimic the behavior of given URL.

The queried URL must be provided with an hash value of this URL that was computed by Idunn before being sent to the user. This intends to avoid potential abusive use of this endpoint.

To small design decisions that may need to be discussed:

 - I don't think Idunn is currently self aware of its public URL, I think this is typically solved by adding a setting. Here I chose to only provide the URL to PJ and the of this URL so that the front-end can build the request to Idunn. It may not be much extensive as we already intended to add some metadata to these URLs by specifying a "label" to display.

 - The `url` and `hash` parameters are passed as traditional query parameters (`URL?param1=val1&param2=val2&...`). I chose this form to avoid having to escape slashes of the parameter URL. However maybe we prefer to stick to a path semantic like `IDUNN_URL/v1/url/{param_url}/{param_hash}.


idunn/api/urlsolver.py
```diff
@@ -0,0 +1,34 @@
...
+# Randomized seed used to hash addresses to solve, do not leek.
+random_token = token_bytes(64)
+
+
+def hash_url(url: str) -> str:
+    """
+    Hash of the URL that the client must provide in order to avoid abusive use
+    of the endpoint.
+    """
+    return sha1(random_token + url.encode()).hexdigest()
```
amatissart, line 21: sha1 is no longer considered safe for cryptographic use, so I would use sha256 instead.

Also, Python provides [`hmac.HMAC` ](https://docs.python.org/3.8/library/hmac.html#hmac.HMAC.hexdigest) to achieve this in a normalized way:
```suggestion
    return hmac.HMAC(key=random_token, msg=url.encode(), digestmod="sha256").hexdigest()
```

idunn/blocks/website.py
```diff
@@ -1,12 +1,21 @@
...
+    redirect: bool = False
+    hash: Optional[str]
+    label: Optional[str]
 
 
 class WebSiteBlock(BaseBlock):
     BLOCK_TYPE: ClassVar = "website"
 
-    url: str
+    url: Union[str, WebsiteUrl]
```
amatissart, line 18: I think we should avoid exposing this complexity to the client.   
As you suggested, I would prefer to set a public base URL in Idunn settings, and use it directly to provide a redirected `url` when needed.
Then, the optional `label` could be added in the existing block (and the block format would not change).

idunn/api/urlsolver.py
```diff
@@ -0,0 +1,34 @@
...
+    if hash_url(url) != hash:
+        raise HTTPException(403, detail="provided hash does not match the URL")
+
+    response = await client.get(url, allow_redirects=False)
+    response.raise_for_status()
+
+    if response.status_code not in range(300, 400):
+        raise HTTPException(404, detail="provided URL does not redirect")
+
+    return RedirectResponse(response.headers["Location"])
```
amatissart, line 34: :+1: looks good to me

idunn/blocks/website.py
remi-dupre: Ok, I went for the solution that required no modification in our configurations by default but I wasn't sure at all, it will be easy to revert :+1:

idunn/api/urlsolver.py
remi-dupre: I feel like a dinosaur ^^'

The doc also recommends to use [`compare_digest`](https://docs.python.org/3.8/library/hmac.html#hmac.compare_digest) to check provided hash. I don't see any reasonable reason why someone would make the effort of putting together a timing attack just to open a single URL, but given the tiny cost I suppose that we're better safe than sorry :D

remi-dupre:
I addressed all your comments, which ends up with:

  - a new `BASE_URL` setting
  - a new `SECRET` setting
  - a new optional field `label` in address blocks

tests/test_pj_poi.py
```diff
@@ -77,7 +77,9 @@ def test_pj_place(enable_pj_source):
     assert blocks[2]["blocks"][0]["blocks"][0]["wheelchair"] == "yes"
 
     assert blocks[3]["type"] == "website"
-    assert blocks[3]["url"] == "http://www.museepicassoparis.fr"
+    assert blocks[3]["url"] == "http://www.museepicassoparis.fr" or blocks[3]["url"].startswith(
+        "http://localhost:5000/v1/url?url=http://www.museepicassoparis.fr&hash=b6fc09"
```
amatissart, line 81: ìt should be `/v1/redirect` right ?

idunn/api/urlsolver.py
```diff
@@ -0,0 +1,42 @@
...
+base_url = settings.get("BASE_URL")
+secret = settings.get("SECRET").encode()
+
+
+def resolve_url(url: str) -> str:
+    """
+    Idunn's URL that can be provided to redirect to the same page as the input
+    URL would.
+    """
+    return base_url + f"v1/redirect?url={quote(url)}&hash={hash_url(url)}"
```
amatissart, line 21: `quote` does not encode `/`, because it's "is intended for quoting the path section of a URL" by default. 
I think that it would be safer to quote them here.
```suggestion
    return base_url + f"v1/redirect?url={quote(url, safe='')}&hash={hash_url(url)}"
```

remi-dupre: Yep, it doesn't cost anything to be careful here

amatissart (approved):

PR #178, opened 2020-12-09
GitHub actions

remi-dupre:
Replaces Travis with Github actions.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          python-version: '3.8'
+
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+
+      - name: Install dependancies
+        run: |
+          pip install pipenv
+          pipenv install --dev
```
amatissart, line 23: I would keep  the `--deploy` option to validate and use the `Pipfile.lock` file to install dependencies
```suggestion
          pipenv install --dev --deploy
```

remi-dupre: That's probably the only think that I could have forgotten ... --'

amatissart: you force-pushed, but no change is visible :slightly_smiling_face:

remi-dupre: Well, that's interesting ^^'

amatissart (approved):

PR #179, opened 2020-12-21
nlu_client: log decisions to skip NLU

remi-dupre:
Add some detail on the reason why a request to the nlu client is aborted. 

This is a simple implementation that will just raise an exception holding a human-readable explanation.

### DIFF
--- a/idunn/api/geocoder.py
+++ b/idunn/api/geocoder.py
@@ -34,6 +34,12 @@ def post_filter_intention(intention, bragi_response, limit):
             matches += 1
             if matches > expected_matches:
                 return True
+
+    logging.info(
+        "Ignored intention which doesn't match enough geocoding results `%s`",
+        intention.description.query,
+        extra={"intention": intention, "bragi_response": bragi_response, "limit": limit},
+    )
     return False
 
 
@@ -54,7 +60,7 @@ async def get_autocomplete(
         try:
             return await nlu_client.get_intentions(text=query.q, lang=query.lang, focus=focus)
         except NluClientException as exp:
-            logger.warning("Ignored NLU for '%s': %s", query.q, exp.reason, extra=exp.extra)
+            logger.info("Ignored NLU for '%s': %s", query.q, exp.reason(), extra=exp.extra)
             return []
 
     autocomplete_response, intentions = await asyncio.gather(
--- a/idunn/api/geocoder.py
+++ b/idunn/api/geocoder.py
@@ -25,8 +25,8 @@ def post_filter_intention(intention, bragi_response, limit):
     # results among geocoded features seem to be too rare.
     expected_matches = limit / 2
     matches = 0
-    for f in bragi_response["features"]:
-        geocoding = f.get("properties", {}).get("geocoding", {})
+    geocodings = [f.get("properties", {}).get("geocoding", {}) for f in bragi_response["features"]]
+    for geocoding in geocodings:
         if geocoding.get("type") != "poi":
             continue
         name = geocoding.get("name", "")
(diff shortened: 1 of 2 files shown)


idunn/api/geocoder.py
```diff
@@ -49,7 +51,11 @@ async def get_intentions():
         if query.lon and query.lat:
             focus = Point(query.lon, query.lat)
 
-        return await nlu_client.get_intentions(text=query.q, lang=query.lang, focus=focus)
+        try:
+            return await nlu_client.get_intentions(text=query.q, lang=query.lang, focus=focus)
+        except NluClientException as exp:
+            logger.warning("Ignored NLU for '%s': %s", query.q, exp.reason, extra=exp.extra)
```
amatissart, line 57: It would be useful to set `reason` in the `extra` object to make logs filtering easier.

Also, I would use `logger.info` instead of `logger.warning` here as this log message will be used for a large portion of requests, and does not draw attention about a specific problem.

idunn/geocoder/nlu_client.py
```diff
@@ -237,9 +242,13 @@ async def get_intentions(self, text, lang, focus=None):
...
 
-        # If there is a label for both a category and a brand, we consider that the request is
-        # ambiguous and ignore both.
-        if bool(brand_query) ^ bool(cat_query):
+        try:
+            if brand_query and cat_query:
+                raise NluClientException("detected a category and a brand")
+
+            if not brand_query and not cat_query:
+                raise NluClientException("no category or brand detected")
```
amatissart, line 250: This is a very common case, for cities, addresses etc.  
Are we sure that we want to include it in logs ?

idunn/api/geocoder.py
```diff
@@ -49,7 +51,11 @@ async def get_intentions():
...
+            return []
```
amatissart, line 58: The new log message does not cover the case when the intention is excluded by `post_filter_intention`. It's probably worth logging them too, as they are intentions we considered irrelevant based on related geocoding results.

idunn/geocoder/nlu_client.py
remi-dupre: I'm not exactly sure, this is indeed not an unexpected case (so info log level sounds fine) but it may be useful to extract some dataset out of the logs? I wouldn't making it more quiet if you think it is better to keep more concise logs :+1:

idunn/geocoder/nlu_client.py
```diff
@@ -289,10 +296,10 @@ async def get_intentions(self, text, lang, focus=None):
                 )
 
             logger.info("Detected intentions for '%s'", text, extra=logs_extra)
-        else:
-            logger.info("No intention detected for '%s'", text, extra=logs_extra)
-
-        return intentions
+            return [intention]
+        except NluClientException as exp:
+            exp.extra = logs_extra
```
amatissart, line 301: `exp.extra` is already set and contains "reason". 
```suggestion
            exp.extra.update(logs_extra)
```

idunn/api/geocoder.py
```diff
@@ -32,6 +34,12 @@ def post_filter_intention(intention, bragi_response, limit):
             matches += 1
             if matches > expected_matches:
                 return True
+
+    logging.info(
+        "Ignored intention which doesn't match enough geocoding results `%s`",
+        intention.description.query,
+        extra={"intention": intention, "bragi_response": bragi_response, "limit": limit},
```
amatissart, line 41: `ìntention` is a Pydantic object and would appear as a `__repr__` string, `intention.dict()` should be used instead.

`bragi_response` is very verbose here. Maybe we could keep places "id" and "label" only, for debugging purposes ?

remi-dupre: Nice catch, it indeed seems dumb to dump the whole geocoding result :+1:

amatissart (approved):

PR #182, opened 2021-01-13
Clean up error handling on /places/<id> and place_from_id

amatissart:
Several different exceptions have been defined to handle these errors, but they were difficult to catch properly.
This PR suggests to derive all these errors from a new `IdunnPlaceError`, and fixes some oversights:

* Unknown id for `ApiPjSource` will now raise `PlaceNotFound`. 
As a consequence `/v1/places/<id>` will return a status 404 instead of 500 in this case.

* `/v1/directions` will accept redirected place ids as `origin` or `destination`
(e.g. unknown `addr:` are redirected to their equivalent `latlon:`)

### DIFF
--- a/idunn/api/directions.py
+++ b/idunn/api/directions.py
@@ -1,7 +1,8 @@
 from fastapi import HTTPException, Query, Depends, Request, Response
 
 from idunn import settings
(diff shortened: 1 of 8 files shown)


remi-dupre (approved):
:+1:

PR #183, opened 2021-01-13
Define workflow to build and push docker image

amatissart:
In preparation for migrating to Qwant organization.  
Similar to https://github.com/Qwant/fafnir/pull/59


remi-dupre (commented):
We could eventually check that the image builds fine on all branches and only condition the upload steps?

remi-dupre (approved):

PR #184, opened 2021-01-14
Richer doc

remi-dupre:
Mostly adds more type information about the public endpoints. As a side effect I attempted to get more precise abstractions, noticeably for things that could be represented by enums (with a bit of hacking for categories).

Blocks are now automatically validated by pydantic by adding the type `Literal[name]` to their field `type`, sadly this is not nicely rendered in the documentation, maybe it will get nicer in future fastapi updates, which would help replacing some regexps for some things like `Literal["yes", "no"]`.

Previous behavior when we had a validation error was to [enforce a code `404`](https://github.com/Qwant/idunn/blob/master/app.py#L33-L35) where fastapi would raise a `422` by default. As this makes it inconsistent with the documentation, would you be OK to revert it to the fastapi's default behavior ?

PS: Would you be interested in setting up a default linter configuration (which would run on the CI), I noticed a few tiny fixes (like unused imports) in recent fixes, it could make it easier? My pylint outputs are usually bloated but I started building a specific configuration for Idunn on some old branch.


Pipfile
```diff
@@ -32,6 +32,7 @@ unidecode = "*"
 httpx = ">=0.12.0"
 geopy = "*"
 py-mini-racer = "*"
+geojson-pydantic = "*"
```
remi-dupre, line 35: I allowed myself to add this tiny dependency, the only interest is to include some documentation about what the geojson schema looks like to potential user who would not be used to it.

I'm not sure if this is a good idea though, but it is quite tiny (just contains some pydantic models) so I don't think we have to be afraid of it not being well maintained or anything on this range. I'd willingly remove it if you think it's a bad idea, the provided example can be quite enough ^^

idunn/api/utils.py
```diff
@@ -28,17 +30,88 @@
...
+class CategoryEnum(str):
+    """
+    Methods defining the behavior of the enum `Category` defined bellow.
+    """
+
+    def match_brand(self):
+        return ALL_CATEGORIES[self].get("match_brand", False)
+
+    def pj_filters(self):
+        return ALL_CATEGORIES[self].get("pj_filters")
```
amatissart, line 74: `pj_what` also needs to be defined similarly.  
"idunn/datasources/pages_jaunes.py" is currently using `c["pj_what"]` and is broken.

idunn/api/places_list.py
```diff
@@ -125,19 +92,6 @@ def valid_raw_filter(cls, v):
...
-        ret = []
-        if not v:
-            return ret
-        for x in v:
-            if x not in ALL_CATEGORIES:
-                raise ValueError(
-                    f"Category '{x}' is invalid since it does not belong to the set of possible categories: {list(ALL_CATEGORIES.keys())}"
-                )
-            ret.append(ALL_CATEGORIES.get(x))
-        return ret
```
amatissart, line 139: `pj_source.get_places_bbox` expects `category` to be a list (possibly empty), and breaks with `categories=None`.

tests/test_happy_hours.py
```diff
@@ -32,52 +32,53 @@ def test_happy_hour():
...
+        is_24_7=False,
         raw="Mo-Sa 20:00-22:00; Su 12:00-14:00, 20:00-22:00",
         days=[
             {
                 "dayofweek": 1,
                 "local_date": "2018-06-11",
-                "status": "yes",
-                "happy_hours": [{"beginning": "20:00", "end": "22:00"}],
+                "status": "open",
+                "opening_hours": [{"beginning": "20:00", "end": "22:00"}],
```
amatissart, line 45: As discussed, I am okay with a breaking change as this block is not currently used.

But does it really make sense to talk about happy hours as "open" and "closed" ?

amatissart (commented):
Very nice to get rid of this hackish `BlocksValidator` :+1: 

There are a few issues with the current usage of `verbosity` and `category` parameters though.

tests/test_happy_hours.py
remi-dupre: Well, that's a good point, I suppose I got lost in the abstractions... ^^

amatissart: If that helps with the maintenance, we could simply delete this block for now, and we'll agree on its format when its use case will be clearer.

remi-dupre: I went for this solution. It seems quite lazy but I suppose it wouldn't be a good idea to keep maintaining unused stuff by default either :|

idunn/api/places_list.py
amatissart: A small issue remained in the `/v1/instant_answer` implementation that passed `category=None` in some cases, where tests are lacking, my bad.

I pushed a fix on this branch, with an additional test case.

amatissart (approved):

PR #186, opened 2021-01-19
Define maps_place_url and maps_directions_url in place 'meta'

amatissart:
Helper functions to build Qwant Maps URLs (implemented by Erdapfel)
are defined in `idunn/utils/maps_urls.py`


remi-dupre (approved):
:+1:

PR #191, opened 2021-02-01
Minor tweaks in PJ datasource

amatissart:
* Remove URL replacement, now that `next_page_url` is correct in the response
* Add an extra margin in `max` parameter to avoid successive queries
* Remove extra whitespace in `what` parameter

### DIFF
--- a/idunn/datasources/pages_jaunes.py
+++ b/idunn/datasources/pages_jaunes.py
@@ -134,8 +134,7 @@ class ApiPjSource(PjSource):
             and res.context.pages.next_page_url
         ):
             pois += self.get_places_from_url(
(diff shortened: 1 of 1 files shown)


idunn/datasources/pages_jaunes.py
```diff
@@ -134,8 +134,7 @@ def get_places_from_url(self, url, params=None, size=10):
             and res.context.pages.next_page_url
         ):
             pois += self.get_places_from_url(
-                res.context.pages.next_page_url.replace("/pros/find", "/pros/search"),
```
remi-dupre, line 137: Oh nice, I also just noticed that this was fixed :)

remi-dupre (approved):

PR #192, opened 2021-02-02
optim: Return ORJSONResponse directly for instant_answer

amatissart:
See https://fastapi.tiangolo.com/advanced/custom-response/?h=orjson#use-orjsonresponse   
and https://fastapi.tiangolo.com/advanced/response-directly/ 
for more details about this change.

This also prevents a double validation after `InstantAnswerResponse` has been created in `build_response`. By default, FastAPI would re-validate the response model when creating the JSON response.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -46,6 +46,7 @@ httpx = ">=0.12.0"
 geopy = "*"
 py-mini-racer = "*"
 geojson-pydantic = "*"
+orjson = "*"
 
 [requires]
 python_version = "3.8"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "21aa8481bd9cd1e43f33cf25051a15e1cec099a5b724252df8074c05980da3ce"
+            "sha256": "ff26f629a8ba907e08845a82af40181c7e0d10de9741bf5a7601c686786d229c"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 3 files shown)


remi-dupre (dismissed):
How impactful is it to our performances? I see in `orjson`'s documentation that it can take ~0.3 ms for the standard library to serialize a ~50kB document (vs. ~0.06 ms for `orjson`), I didn't expect that to be so slow O_o

amatissart:
> How impactful is it to our performances? I see in `orjson`'s documentation that it can take ~0.3 ms for the standard library to serialize a ~50kB document (vs. ~0.06 ms for `orjson`), I didn't expect that to be so slow O_o

Here is what I could measure on my local machine (with cached responses from NLU and PJ, to avoid fluctuation)

**Before this PR**
```
 %timeit -r10 -n1 session.get('http://localhost:5000/v1/instant_answer?q=restaurant+à+paris&lang=fr')
84.6 ms ± 7.78 ms per loop (mean ± std. dev. of 10 runs, 1 loop each)
```

**With a direct `ORJSONResponse`**
```
 %timeit -r10 -n1 session.get('http://localhost:5000/v1/instant_answer?q=restaurant+à+paris&lang=fr')
60.7 ms ± 11.4 ms per loop (mean ± std. dev. of 10 runs, 1 loop each)
```

Not huge, but measurable.  
The difference is mostly due to the second validation Pydantic models that is now skipped by returning a direct response, and not the serialization itself. So using the default `JSONResponse` as a direct response would lead to very similar results.
Nevertheless I think it's still interesting to integrate "orjson", as it's a very easy change and could be significant for bigger payloads (geographical boundaries, etc.).

Also note that the remaining response time (except network calls), is still mostly spent in opening hours evaluation: ~ 5ms per block.

remi-dupre (approved):

remi-dupre:
> Also note that the remaining response time (except network calls), is still mostly spent in opening hours evaluation: ~ 5ms per block.

I'll try to run the same test with a rebased version of https://github.com/Qwant/idunn/pull/176, I wonder how my current compiled implementation performs on this real world usage :)

PR #195, opened 2021-02-08
instant answer: more permissive filter

remi-dupre:
Add a richer filter for results, among things that remain to do:

I didn't attempt to considering the order of the terms as I think it would be much more complicated to filter out only a few specific cases, is this fair?

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -142,6 +142,7 @@ disable=print-statement,
 # Not disabled in default config
         broad-except,
         consider-using-get,
+        dangerous-default-value,
         global-statement,
         missing-class-docstring,
         missing-function-docstring,
--- a/idunn/utils/result_filter.py
+++ b/idunn/utils/result_filter.py
@@ -51,7 +51,7 @@ def word_as_number(word):
         return word
 
     for suffix in NUM_SUFFIXES:
-        if word.lower().endswith(suffix):
+        if word.endswith(suffix):
             prefix = word[: -len(suffix)]
 
             if prefix.isnumeric():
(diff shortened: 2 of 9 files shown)


.pylintrc
```diff
@@ -142,6 +142,7 @@ disable=print-statement,
 # Not disabled in default config
         broad-except,
         consider-using-get,
+        dangerous-default-value,
```
remi-dupre, line 145: FYI it was complaining about default list initialization:

```python
# If `f` mutates `vals` this may lead to undesired behavior.
def f(vals: List = []):
    pass
```

This is an interesting lint but let's stick to our decision of avoiding pain by default ^^

idunn/utils/result_filter.py
```diff
@@ -0,0 +1,183 @@
...
+
+def words(text):
+    """
+    Split words into a list.
+
+    >>> words("17, rue jaune ; Levallois-Peret")
+    ['17', 'rue', 'jaune', 'Levallois', 'Peret']
+    """
+    separators = [" ", "-", ",", ";", "\\.", "'"]
+    return [word for word in re.split("|".join(separators), text) if word]
```
amatissart, line 32: There are many redundant calls to `.lower()` in this file. Maybe we could return lowercase words here ?

idunn/utils/result_filter.py
```diff
@@ -0,0 +1,183 @@
...
+# Typical suffixes found after numbers such as "4bis", "4th", ...
+NUM_SUFFIXES = ["bis", "ter", "quad", "e", "è", "eme", "ème", "er", "st", "nd", "rd", "th"]
+
+# Abrevation typical found in places, addresses, ...
+ABREVIATIONS = {
+    "av": "avenue",
+    "bd": "boulevard",
+    "r": "rue",
+    "st": "saint",
+}
```
amatissart, line 21: A few more items could be imported from https://github.com/CanalTP/mimirsbrunn/blob/master/libs/mimir/src/rubber.rs#L55-L60

amatissart (approved):

PR #197, opened 2021-02-10
Instant answer: normalization by removing common patterns in queries

amatissart:
This PR introduces
* a "patterns_to_ignore.txt " file, with regex to remove at the start and end of queries for instant answers
* a new flag `IA_SUCCESS_ON_GENERIC_QUERIES`, to return a successful response with empty results on queries with no substantial input. This is useful when the client would like to handle that case specifically, with a generic map for example.

### DIFF
--- a/idunn/instant_answer/patterns_to_ignore.txt
+++ b/idunn/instant_answer/patterns_to_ignore.txt
@@ -1,0 +1,25 @@
+qwant ?maps?
+maps? qwant
+google ?maps?
+maps? google
+maps?( of)?
+carte( des?| du)?
+plan( de| du)?
+add?resse?
+((num(e|é)ro)|n(o|°)?)( de)? ?t(e|é)l(e|é)phone
+num(e|é)ro$
+t(e|é)l(e|é)phone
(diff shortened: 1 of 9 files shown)


idunn/instant_answer/patterns_to_ignore.txt
```diff
@@ -0,0 +1,25 @@
+qwant ?maps?
+maps? qwant
+google ?maps?
+maps? google
+maps?( of)?
+carte( des?| du)?
+plan( de| du)?
+add?resse?
+((num(e|é)ro)|n(o|°)?)( de)? ?t(e|é)l(e|é)phone
+num(e|é)ro$
```
remi-dupre, line 10: Is `$` intentional? You may have an example in mind where it makes sense to keep this keyword?

```suggestion
num(e|é)ro
```

(same with `menu$` below)

amatissart: These tokens are mostly used at the end of queries, so this was a (superfluous ?) precaution for cases when a specific POI name would start with "numero ..." or "menu ...".

remi-dupre (approved):

PR #201, opened 2021-02-18
Define source_url and contribute_url in place meta

amatissart:
In the `meta` object:
* `source_url` is a direct URL to the place details
* `contribute_url` is defined for OSM POIs only

### DIFF
--- a/idunn/places/pj_poi.py
+++ b/idunn/places/pj_poi.py
@@ -220,6 +220,12 @@ class PjPOI(BasePlace):
     def get_source(self):
         return PoiSource.PAGESJAUNES
 
+    def get_source_url(self):
+        business_id = self.get("BusinessId")
+        if not business_id:
+            return None
+        return f"https://www.pagesjaunes.fr/pros/{business_id}"
+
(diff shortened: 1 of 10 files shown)


idunn/places/pj_poi.py
```diff
@@ -388,6 +394,16 @@ def get_images_urls(self):
     def get_source(self):
         return PoiSource.PAGESJAUNES
 
+    def get_source_url(self):
+        return next(
+            (
+                ins.urls.merchant_url
```
remi-dupre, line 400: We may want to avoid exposing users to some stats?

```suggestion
                resolve_url(ins.urls.merchant_url)
```

They should end up on pagesjaunes.fr anyway so there may not be much more information that could be fetched. (Maybe if we want to be extra safe we could check that the link actually redirects to pagejaunes.fr?)

bbecquet (approved):
Thanks!
As the design for IA were simplified lately, we'll only use `source_url` on them. But `contribute_url` will be great to simplify Erdapfel also :+1:

PR #204, opened 2021-02-25
result_filter: check translated names

remi-dupre:
Read from translation fields while filtering instant answer's output. This should mainly help when the content of the query is not consistent with parameter `lang`.

### DIFF
--- a/idunn/utils/result_filter.py
+++ b/idunn/utils/result_filter.py
@@ -112,13 +112,13 @@ def word_matches(query_word, label_word):
 
 def check(
     query: str,
-    name: str,
+    names: List[str],
     admins: List[str] = [],
     postcodes: List[str] = [],
     is_address: bool = True,
 ) -> bool:
     query_words = words(query.lower())
-    name_words = words(name.lower())
+    names = list(map(str.lower, names))
     admins = list(map(str.lower, admins))
 
     if is_address:
@@ -126,7 +126,7 @@ def check(
 
     # Check if all words of the query match a word in the result
     full_label = [
-        *name_words,
+        *(w for name in names for w in words(name)),
         *postcodes,
         *(w for admin in admins for w in words(admin)),
     ]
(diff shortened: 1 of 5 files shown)


idunn/utils/result_filter.py
```diff
@@ -135,19 +135,23 @@ def check(
...
+    #   - over one of the result names together with a postcode
     def coverage(terms):
         return sum(
             any(word_matches(query_word, term) for query_word in query_words) for term in terms
         ) / len(terms)
 
     check_coverage = any(
         coverage(terms) >= (2 / 3)
+        for name_words in map(words, names)
         for terms in [
```
remi-dupre, line 155: Looking at the Eiffel Tower, in the worst cases there will be ~50 translations, ~10 admins which leads to about 1500 calls to `coverage` (which is about quadratic in the number of words of the query, that shouldn't be too expensive).

I don't think it will be an issue but if it is, it could be relevant to take another approach, like checking only a few languages (like French + English + `lang` + local name)

amatissart: A simple way to attenuate this overhead would be to use a lru_cache on `word_matches`, as common words are likely to appear in different languages:

```diff
+@lru_cache(maxsize=10000)
 def word_matches(query_word, label_word):
     """
     Check if two words match with an high degree of confidence.
```

idunn/utils/result_filter.py
```diff
@@ -180,9 +184,16 @@ def check_bragi_response(query: str, bragi_response: dict) -> bool:
             }
         )
 
+    name = bragi_response["name"]
+    local_names = [
+        prop["value"]
+        for prop in bragi_response.get("properties", [])
```
amatissart, line 190: Side note: these properties are defined for POIs only.  
The issue remains for administrative regions, as Bragi returns a single name for these objects (according to the requested language). Let's ignore this for now, as it would require either to update the Bragi response, or an additional request to fetch the object from Elasticsearch.

idunn/utils/result_filter.py
remi-dupre: Well an extra dict lookup is almost free so I'll include this change. I don't think this will become an issue anyway

idunn/utils/result_filter.py
remi-dupre: So the issue will kind of persist. Do you know if there is a motivation to not include these properties for admins in bragi ? It would actually be quite cheap to just add a query parameter that would disable this behavior ?

amatissart: In mimir objects, contrary to `Poi`, [`Admin`](https://github.com/CanalTP/mimirsbrunn/blob/master/libs/mimir/src/objects.rs#L516) does not contain any `properties`. Localized names are only stored in `names`. And for most geocoding use-cases, these names are not useful in the response.  

It's true that providing `properties` for all objects would be more consistent, at least as an option. But it's unclear for me whether these properties should be derived from the existing `names` field, or be added into the data model for `Admin` (and be provided at import time).

amatissart (approved):

PR #205, opened 2021-02-25
result_filter: re-rank results when query order is not be reflected

remi-dupre:
~~(based on #204)~~

### DIFF
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -127,16 +127,30 @@ async def get_instant_answer(
 
     if not intentions:
         bragi_response = await bragi_client.raw_autocomplete(
-            {"q": normalized_query, "lang": lang, "limit": 1}
+            {"q": normalized_query, "lang": lang, "limit": 5}
         )
-        if len(bragi_response["features"]) > 0:
-            place_geocoding = bragi_response["features"][0]["properties"]["geocoding"]
-            place_id = place_geocoding["id"]
-            if result_filter.check_bragi_response(normalized_query, place_geocoding):
-                result = await run_in_threadpool(
-                    get_instant_answer_single_place, place_id=place_id, lang=lang
-                )
-                return build_response(result, query=q, lang=lang)
+
+        geocodings = sorted(
+            (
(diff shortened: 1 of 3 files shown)


idunn/api/instant_answer.py
```diff
@@ -127,16 +127,30 @@ async def get_instant_answer(
...
+                for feature in bragi_response["features"]
+                for rank in [
+                    result_filter.rank_bragi_response(
+                        normalized_query, feature["properties"]["geocoding"]
+                    )
+                ]
+                if rank is not None
+            ),
+            key=lambda item: -item[0],  # sort by descending rank
+        )
```
amatissart, line 145: As discussed, this `rank_bragi_response` may take several milliseconds in some cases.  
I would suggest to ignore this reranking step when the first result is neither a street or an address, as the first result will not change anyway.

tests/test_result_filter.py
```diff
@@ -46,7 +46,28 @@ def test_filter():
...
+    }
+
+    paris = {
+        "names": ["rue de Rennes"],
+        "admins": ["Paris"],
+        "place_type": "street",
+    }
+
+    assert rank("rue de Paris, Rennes", **rennes) > rank("rue de Rennes, Paris", **rennes)
+    assert rank("rue de Rennes, Paris", **paris) > rank("rue de Paris, Rennes", **paris)
```
amatissart, line 73: If i am not mistaken, it would be more interesting to compare rank values for the same query:
```py
    assert rank("rue de Rennes, Paris", **paris) > rank("rue de Rennes, Paris", **rennes)
    assert rank("rue de Paris, Rennes", **rennes) > rank("rue de Paris, Rennes", **paris)
```

amatissart (commented):
Looks good to me, except a comment about tests

tests/test_result_filter.py
remi-dupre: Nice catch! I'm quite embarrassed as I actually though about it and still made the wrong choice ... :facepalm:

amatissart (approved):

PR #206, opened 2021-02-25
update black

remi-dupre:
Update black to latest version.

Most impactful changes appear to be:

 - a bit of normalization in the docstrings
 - trailing commas enforce to split into one line per item (apparently this is called [magic trailing](https://github.com/psf/black/pull/1612))

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "86016b37eb9e920c40d1dc1767d63b9e695a4ab66ec547315d541ce7d4040838"
+            "sha256": "248fc08d5aabfd54d3abbbb9c93db3174899bfa65b47fbe12ce69a5d4f70b784"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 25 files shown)


Pipfile.lock
```diff
@@ -24,6 +31,13 @@
             "index": "pypi",
             "version": "==2.9.0"
         },
+        "black": {
+            "hashes": [
+                "sha256:1c02557aa099101b9d21496f8a914e9ed2222ef70336404eeeac8edba836fbea"
+            ],
+            "index": "pypi",
+            "version": "==20.8b1"
+        },
```
amatissart, line 40: I am confused about why "black" and its dependencies are added here, whereas they are already defined as dev packages. These new definitions are also removed from Pipfile.lock after running `pipenv lock`, so I don't think they are supposed to be committed here.

remi-dupre: I must have done something wrong indeed!

I often feel like an old lady when working with `pipenv`, this must be some kind of curse because it is not a very complicated tool :'D

remi-dupre: This is fixed, I just ran a `pipenv install --dev --selective-upgrade "black==20.8b1"` from master.

I think that I initially forgot the `--dev` yesterday, then I somehow failed or forgot to revert the change.

remi-dupre: Aaaaand also required a `pipenv lock --keep-outdated` which added this hash: https://github.com/Qwant/idunn/pull/206/files#diff-a86c67a0a29ed0e95909b9b7c420140f302d17399ee6dcce4e1a51a14d27fd51R499.

amatissart (approved):

PR #207, opened 2021-02-26
PJ API: Remove places with duplicated merchant_id from results

amatissart:
It has been noticed that different result pages from the API may contain duplicated POIs.

This PR suggests to remove these duplicated items. As a result, the number of places in the response, may be strictly less than the requested "size", even if more places would have been available in following pages.

### DIFF
--- a/idunn/datasources/pages_jaunes.py
+++ b/idunn/datasources/pages_jaunes.py
@@ -160,7 +160,18 @@ class ApiPjSource(PjSource):
             query_params["what"] += " " + query
             query_params["what"] = query_params["what"].strip()
 
-        return self.get_places_from_url(self.PJ_FIND_API_URL, query_params, size)
+        api_places = self.get_places_from_url(self.PJ_FIND_API_URL, query_params, size)
+
+        # Remove duplicated merchant ids that may be returned in different pages
+        merchant_ids = set()
+        places = []
+        for p in api_places:
+            merchant_id = p.data.merchant_id
+            if merchant_id not in merchant_ids:
+                places.append(p)
+                if merchant_id:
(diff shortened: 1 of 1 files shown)


bbecquet (approved):

idunn/datasources/pages_jaunes.py
```diff
@@ -160,7 +160,18 @@ def get_places_bbox(
...
+        # Remove duplicated merchant ids that may be returned in different pages
+        merchant_ids = set()
+        places = []
+        for p in api_places:
+            merchant_id = p.data.merchant_id
+            if merchant_id not in merchant_ids:
+                places.append(p)
+                if merchant_id:
+                    merchant_ids.add(merchant_id)
+        return places
```
remi-dupre, line 174: I was wondering if there was a more declarative way to do it. I don't really think it is better (especially with black enforcing it as a one-liner) but still funny:

```suggestion
        return list({place.data.merchant_id: place for place in api_places}.values())
```

remi-dupre (approved):

idunn/datasources/pages_jaunes.py
bbecquet: I was looking for something similar but discarded it because I thought the order was lost in a dictionary… but apparently since Python 3.7 it is not.

Actually I hoped to find something as elegant as [Ramda.uniqBy](https://ramdajs.com/docs/#uniqBy) :)

amatissart: I agree that writing 9 lines for this does not look satisfying. But the one-liner would require a double `reversed` to keep the first occurrence, and it also collapses `None` keys (they are not supposed to exist, but are not duplicates per se).

PR #208, opened 2021-03-01
Minor adjustments on instant answers triggers and normalization

amatissart:
After internal feedbacks and observations of my own.

### DIFF
--- a/idunn/instant_answer/patterns_to_ignore.txt
+++ b/idunn/instant_answer/patterns_to_ignore.txt
@@ -23,3 +23,4 @@ lieu dit
 localisation
 o(ù|u) (se|ce) (trouve|situe)( (le|la))?
 where is
+booking$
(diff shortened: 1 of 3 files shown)


remi-dupre (approved):

idunn/instant_answer/patterns_to_ignore.txt
```diff
@@ -23,3 +23,4 @@ lieu dit
 localisation
 o(ù|u) (se|ce) (trouve|situe)( (le|la))?
 where is
+booking$
```
bbecquet, line 26: Why did we choose to match only at the end of the string?

amatissart: To reduce the risk of false positives, when the keyword may be helpful to recognize and classify other types of intention.  
When associated to a hotel name, "booking" appears almost always at the end of the query (statistically).

PR #209, opened 2021-03-04
Anti scraping integration

amatissart:
For instant answers only at this point:
 * Integrate "ban redis" mechanism: return 429 status code for clients found in redis
 * Limit query length (to 100 chars by default)

### DIFF
--- a/idunn/api/urls.py
+++ b/idunn/api/urls.py
@@ -1,3 +1,5 @@
+from fastapi import Depends
+
 from .pois import get_poi
 from .places import get_place, get_place_latlon
 from .status import get_status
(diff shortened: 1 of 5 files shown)


idunn/utils/default_settings.yaml
```diff
@@ -177,4 +177,13 @@ RECYCLING_MEASURES_MAX_AGE_IN_HOURS: 168 # 7 days (Older measures will be ignore
 
 ##########################
 ## Instant Answer
+IA_MAX_QUERY_LENGTH: 100
```
remi-dupre, line 180: Should we consider extending it to non IA queries? I'm especially thinking about direct queries (`?q=...`)

idunn/api/urls.py
```diff
@@ -85,6 +87,7 @@ def get_api_urls(settings):
         APIRoute(
             "/instant_answer",
             get_instant_answer,
+            dependencies=[Depends(check_banredis)],
```
remi-dupre, line 90: That looks great!

remi-dupre (approved):

remi-dupre (approved):

idunn/utils/default_settings.yaml
amatissart: Possibly. But direct queries to /maps/?q are forwarded to "/v1/autocomplete" by Erdapfel. So currently, there is nothing we can do in Idunn to handle them differently.

PR #211, opened 2021-03-16
Monitoring and logging adjustments for instant answers

amatissart:
Each commit can be reviewed independently

- count and time all requests in Prometheus metrics, including `HTTPException` with error status codes, and other exceptions
- update "result_filter" log level
- add log messages about  instant answer responses, to be able to analyze the proportion of valid responses based on these logs

### DIFF
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -71,19 +71,18 @@ class InstantAnswerResponse(BaseModel):
     data: InstantAnswerData
 
 
-class NoInstantAnswerToDisplay(HTTPException):
-    def __init__(self, query=None, lang=None):
-        super().__init__(status_code=404, detail="No instant answer to display")
-        if query is not None:
-            logger.info(
-                "get_instant_answer: no answer",
-                extra={
-                    "request": {
-                        "query": query,
-                        "lang": lang,
-                    }
-                },
-            )
+def no_instant_answer(query=None, lang=None):
+    if query is not None:
+        logger.info(
+            "get_instant_answer: no answer",
+            extra={
+                "request": {
+                    "query": query,
+                    "lang": lang,
+                }
+            },
+        )
+    raise HTTPException(status_code=404, detail="No instant answer to display")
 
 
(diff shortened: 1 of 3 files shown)


idunn/api/instant_answer.py
```diff
@@ -71,10 +71,36 @@ class InstantAnswerResponse(BaseModel):
...
+        if query is not None:
+            logger.info(
+                "get_instant_answer: no answer",
+                extra={
+                    "request": {
+                        "query": query,
+                        "lang": lang,
+                    }
+                },
+            )
```
remi-dupre, line 86: Isn't it a bit odd to log from the exception's constructor? `query` and `lang` are not actually part of the exception data, that feels a bit confusing.

I'm not sure, but maybe raising the exception from an explicit function or method would make more sense ?

amatissart: You're right. I was a bit reluctant to change how exception was raised, but logging from the constructor is probably not a good pattern.

The exception is now raised from a function `no_instant_answer` is defined. This led me to to some refactoring: "get_instant_answer" was a very long function, and the linter complained about too many `return`. 

So it should be more readable now. What do you think ?

idunn/api/instant_answer.py
```diff
@@ -99,68 +126,9 @@ def get_instant_answer_single_place(place_id: str, lang: str):
...
-            )
-            return build_response(result, query=q, lang=lang)
-
-        raise NoInstantAnswerToDisplay
-
-    intention = intentions[0]
+async def get_instant_answer_intention(intention, query: str, lang: str):
     if not intention.filter.bbox:
-        raise NoInstantAnswerToDisplay
+        return no_instant_answer(query=query, lang=lang)
```
remi-dupre, line 131: I think I would have called the function `raise_no_instant_answer` instead, it would describe the intention slightly better. Here the `return` could lead to think that a value will actually be returned.

This is mostly subjective again so I wouldn't be surprised if you don't agree ^^

remi-dupre (commented):
Sorry for the late review :confused:

remi-dupre (approved):

PR #212, opened 2021-03-18
instant_answer: use focus when a region is specified

remi-dupre:
Use a region code in the params to trigger use of the focus in instant answers.

For example when looking for "Ikea la vallette" from france:

```bash
# We get results from Germany by default
$ curl "http://localhost:5000/v1/instant_answer?q=ikea%20la%20valette&lang=fr" | grep "Am Wandersmann 2-4 (Hofheim am Taunus)" | wc -l
1
# We get results from France with the region set
$ curl "http://localhost:5000/v1/instant_answer?q=ikea%20la%20valette&lang=fr&user_region=fr" | grep "zac Valgora, 83160 LA VALETTE DU VAR" | wc -l
1
```

This still requires:

 - [x] stabilization of the parameter name
 - [x] a few tests, there is actually two distinct mechanisms using the focus (for intentions and no intentions)

I used the region database from [erdapfel](https://github.com/Qwant/erdapfel/blob/master/config/regions.yaml) and removed fields that are not likely to be used. I kept `bbox` as it may be useful in the future if we want to tweak the zoom level depending on the region size (a purely naive approach would not fit, for example France's bbox covers most of the world).


idunn/geocoder/models/params.py
```diff
@@ -19,8 +18,7 @@
 FOCUS_DECAY = float(settings["FOCUS_DECAY"])
 
 
-@dataclass
-class QueryParams:
+class QueryParams(BaseModel):
```
amatissart, line 21: Unfortunately, this causes unexpected changes about how FastAPI handles the parameters. `Query()` used as a default value in a pydantic model is ignored by FastAPI. As a result, all non-str types are considered as fields that should appear in the request body. And moreover, most of the documentation would be lost.

The good news is that it should be possible to keep most of your changes AND the `@dataclass`.
The `shape` field could indeed be defined with `Body(None,..., embed=True)`
See https://fastapi.tiangolo.com/tutorial/body-multiple-params/#embed-a-single-body-parameter

remi-dupre: Oh right, sorry about that, this is quite embarassing.

My main issue was that with this use of pydantic's dataclasses it is not possible to use the default constructor since the default values are in the form `Query(...)`. However it is possible to use `__pydantic_model__` to access inner model, which handles `Query` correctly :tada: 

This significant change of behavior by using a similar type which actually behaves strangely really seems very counter intuitive to use :disappointed: I suppose this behavior for dataclasses is slightly hacky and unintended ? (A quick search through the docs only raises the changelog from the date it was supported, which feels like it should only be an alternative to the use of models)

remi-dupre:
@amatissart I've just rebased and pushed

amatissart (approved):

PR #215, opened 2021-03-19
pagesjaunes: build affiliation links

remi-dupre:
Build affiliation links instead of using links provided by the API to avoid slight privacy concerns.

The most naive approach would be to provide `https://www.pagesjaunes.fr/pros/{merchant_id}#blocAvis` ~~but we still want to allow some metrics so I kept `https://www.pagesjaunes.fr/partenaire?id={merchant_id}&component=review&action=view&xtor=AD-350-[QWANT]-[CONTRIBUTIF]-[VOIR_TOUS_AVIS]` from production.~~

### DIFF
--- a/idunn/places/pj_poi.py
+++ b/idunn/places/pj_poi.py
@@ -395,14 +395,7 @@ class PjApiPOI(BasePlace):
         return PoiSource.PAGESJAUNES
 
     def get_source_url(self):
-        return next(
-            (
-                ins.urls.merchant_url
-                for ins in self.data.inscriptions
-                if ins.urls and ins.urls.merchant_url
-            ),
-            None,
-        )
+        return f"https://www.pagesjaunes.fr/pros/{self.data.merchant_id}"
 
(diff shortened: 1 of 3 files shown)


idunn/places/pj_poi.py
```diff
@@ -425,11 +418,7 @@ def get_raw_grades(self):
...
-        return next(
-            (
-                ins.urls.reviews_url
-                for ins in self.data.inscriptions
-                if ins.urls and ins.urls.reviews_url
-            ),
-            None,
+        return (
+            f"https://www.pagesjaunes.fr/partenaire?id={self.data.merchant_id}&component=review"
+            "&action=view&xtor=AD-350-[QWANT]-[CONTRIBUTIF]-[VOIR_TOUS_AVIS]"
```
amatissart, line 423: As discussed, a simple URL is also available, based on the API response.

amatissart (approved):

PR #217, opened 2021-03-24
extend the list of categories

remi-dupre:
Extend the list of categories to match what the classifier can identify.

Some updates to karthoterian_docker (WIP) are required to cover all the POIs and categories, especially:

 - restaurant types, I think we can match on the `cuisine` tag prior to `leisure` in order to set `class,subclass` = `restaurant,cuisine_type`.
 - I removed diet-related categories since it would be quite annoying to add this information right now (I have a few ideas but they are all quite hacky)
 - we could match more subkeys for some tags : tourism, historic

Screenshot_2021-04-12_08-49-28

### DIFF
--- a/idunn/utils/categories.yml
+++ b/idunn/utils/categories.yml
@@ -1,144 +1,602 @@
 categories:
-    restaurant:
-        pj_what: "restaurants"
-        pj_filters:
-            - "restaurants"
-        raw_filters:
-            - "restaurant,*"
-            - "fast_food,*"
-        regex: "^resto|^restau"
-    hotel:
-        pj_what: "hôtels"
-        pj_filters:
-            - "hôtels"
-        raw_filters:
-            - "*,hotel"
-        regex: "hotel"
-    cinema:
-        pj_what: "cinéma"
-        pj_filters:
-            - "salles de cinéma"
-        raw_filters:
-            - "cinema,*"
-        regex: "^cine"
     # NOTE: This is a legacy alias of `cinema`.
     #       It may be removed once erdapfel don't rely on it anymore.
     leisure:
         pj_what: "cinéma"
         pj_filters:
-            - "salles de cinéma"
+          - "salles de cinéma"
         raw_filters:
-            - "cinema,*"
-    theatre:
-        pj_what: "théatre"
+          - "cinema,*"
+
+    # NOTE: This is a legacy alias of `health_hospital`.
+    health:
+        raw_filters:
+          - "clinic,*"
+          - "hospital,*"
+        pj_what: "hôpital"
         pj_filters:
-            - "salles de concerts, de spectacles"
-        raw_filters:
-            - "theatre,*"
-        regex: "^theat"
+          - "Hôpital"
+        regex: "health|sante|hopital|hospital|clinic|clinique"
+
+    # NOTE: This is a legacy alias for `health_pharmacy`
     pharmacy:
         pj_what: "pharmacie"
         pj_filters:
-            - "Pharmacie"
+          - "Pharmacie"
         raw_filters:
-            - "pharmacy,*"
+          - "pharmacy,*"
         regex: "pharmac"
+
+    # NOTE: This is a legacy alias for shop_supermarket
     supermarket:
         pj_what: "alimentation"
         pj_filters:
-            - "supermarchés, hypermarchés"
+          - "supermarchés, hypermarchés"
         raw_filters:
-            - "*,supermarket"
-            - "*,mall"
+          - "*,supermarket"
+          - "*,mall"
         regex: "superette|epicier|epicerie|marche|market"
-    bank:
-        pj_what: "banques"
-        pj_filters:
-            - "banques"
-        raw_filters:
-            - "bank,*"
-        regex: "^(bank|banque|credit|caisse|atm$)"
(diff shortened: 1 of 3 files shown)


idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+
+    fuel:
+        raw_filters:
+          - "fuel,*"
+        pj_what: "station essence"
+
+    community_centre:
+        raw_filters:
+          - "*,community_centre"
+        pj_what: "centre communal"
```
amatissart, line 307: Idem, `pj_what` may not be necessary ?

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+        raw_filters:
+          - "camp_site,*"
+          - "caravan_site,*"
+        pj_what: "camping"
+
+    station:
+        regex: "^gare( |$)"
+        raw_filters:
+          - "railway,station"
+          - "railway,tram_stop"
```
amatissart, line 336: Is it really relevant to include "tram_stop" here ? And why not "subway" too in this case ?  
An intention about "station" would probably be about actual train stations, don't you think ?

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+          - "nightclub,*"
+          - "leisure,escape_game"
+          - "leisure,golf_course"
+          - "leisure,miniature_golf"
+          - "tourism,aquarium"
+          - "tourism,artwork"
+          - "tourism,attraction"
+          - "tourism,theme_park"
+          - "tourism,zoo"
+        pj_what: "loisir"
```
amatissart, line 356: Same question about "pj_what". Does that provide relevant results ? And then, could it be used for "leisure" too ?

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
-            - "police,*"
-        regex: "(^| )(police|gendar)"
-    fitness:
-        pj_what: "sport"
+          - "*,optician"
+        pj_what: "opticien"
+
+    grave_yard:
+        raw_filters:
+          - "grave_yard,*"
```
amatissart, line 434: A class "cemetery" exists and should be used instead:
https://github.com/Qwant/openmaptiles/blob/master/layers/poi/class.sql#L55

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+        pj_what: "boutique de chaussures"
+
+    shop_doityourself:
+        raw_filters:
+          - "*,doityourself"
+          - "*,interior_decoration"
+          - "*,hardware"
+        pj_what: "magasin de bricolage"
+
+    tourismebicycle_rental:
```
amatissart, line 464: is there a typo in the category name ?

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+          - "*,post_box"
+
+    playground:
+        raw_filters:
+          - "playground,*"
+
+    marketplace:
+        raw_filters:
+          - "*,marketplace"
+        pj_what: "marché"
```
amatissart, line 598: same question: is PJ a relevant source for that category ?

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+
+    shop_travel_agency:
+        raw_filters:
+          - "*,travel_agency"
+        pj_what: "agence de voyage"
+
+    shop_other:
+        raw_filters:
+          - "shop,*"
+        pj_what: "magasin"
```
amatissart, line 572: Does that really make sense to provide a list of results for such a wide category ?

idunn/utils/categories.yml
```diff
@@ -1,144 +1,602 @@
...
+        pj_what: "restaurant thaï"
+
+    food_vietnamese:
+        raw_filters:
+          - "restaurant,*vietnamese*"
+        pj_what: "restaurant vietnamien"
+
+    food_lebanese:
+        raw_filters:
+          - "restaurant,*lebanese*"
```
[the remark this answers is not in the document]
remi-dupre: It could help to implement this kind of behavior, for example in cases where subclass would contain several comma-separated values (or maybe imposm actually split those?), but it is quite irrelevant now and is mostly an omission in the translation of previously existing raw SQL filters :/

idunn/utils/categories.yml
```diff
@@ -1,144 +1,584 @@
...
+          - "camp_site,*"
+          - "caravan_site,*"
+        pj_what: "camping"
+
+    station:
+        regex: "^gare( |$)"
+        raw_filters:
+          - "railway,station"
+          - "railway,halt"
+        pj_what: "gare"
```
amatissart, line 326: Sorry for the oversight, I think that this `pj_what` should be removed here too. It was not present before.  
Maybe false positives are returned by PJ for such a query.

Apart from that, everything looks good to me :+1:

remi-dupre: That's fixed, sorry about this, my approach to this PR was absolutely messy and counter-productive ... I'm sorry about the time it must have taken to review and quite relieved it comes to an end :|

amatissart (approved):

PR #224, opened 2021-04-23
add endpoint `search`

remi-dupre:
This mostly reuses code from `instant_answer` but I tried to change the API a bit to be more consistent with what we usually use internally between erdapfel and Idunn:

 - the answer format is simpler, either consisting of the same format as `get_place` (through key `place`) or the format of `get_bbox_places` (through key `bbox_places`)
 - we don't usually raise error codes 404 when we have an empty geocoder result, I tried to stick with this by returning an empty result

The only slight behavior change is that there is no truncating of the input query, maybe we should add a global parameter for this? I've also thought about skipping normalization since it would seem weird to "ignore or change part of what is in the search bar" but this seems like a free potential improvement of results?

Part of the logs may be weird with the current state of the branch, I'm still thinking there may be some global pattern that could be built since fetching the context from the logger is a common issue we have (eg. how we build extra infos in `nlu_client` or in this case understanding what endpoint is currently being used).

### DIFF
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -181,7 +181,7 @@ async def get_instant_answer_intention(intention, query: str, lang: str):
             maps_frame_url=maps_urls.get_places_url(intention.filter, no_ui=True),
         )
 
-    return result
+    return build_response(result, query=query, lang=lang)
 
 
 async def get_instant_answer(
@@ -223,11 +223,7 @@ async def get_instant_answer(
                 normalized_query, lang, extra_geocoder_params
             )
             if intentions:
-                return build_response(
-                    await get_instant_answer_intention(intentions[0], query=q, lang=lang),
-                    q,
-                    lang,
-                )
+                return await get_instant_answer_intention(intentions[0], query=q, lang=lang)
         except NluClientException:
             # No intention could be interpreted from query
             pass
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -181,7 +181,7 @@ async def get_instant_answer_intention(intention, query: str, lang: str):
             maps_frame_url=maps_urls.get_places_url(intention.filter, no_ui=True),
         )
 
-    return build_response(result, query=query, lang=lang)
+    return result
 
 
 async def get_instant_answer(
@@ -223,7 +223,11 @@ async def get_instant_answer(
                 normalized_query, lang, extra_geocoder_params
             )
             if intentions:
-                return await get_instant_answer_intention(intentions[0], query=q, lang=lang)
+                return build_response(
+                    await get_instant_answer_intention(intentions[0], query=q, lang=lang),
+                    q,
+                    lang,
+                )
         except NluClientException:
             # No intention could be interpreted from query
             pass
--- a/idunn/api/search.py
+++ b/idunn/api/search.py
@@ -1,0 +1,115 @@
+import logging
+from fastapi import Depends, HTTPException
+from fastapi.concurrency import run_in_threadpool
+from typing import Optional
+from pydantic import BaseModel, Field
+
+from idunn import settings
+from idunn.api.places_list import PlacesBboxResponse
+from idunn.geocoder.bragi_client import bragi_client
+from idunn.geocoder.models.geocodejson import Intention
+from idunn.geocoder.nlu_client import nlu_client, NluClientException
+from idunn.places import place_from_id, Place
+from idunn.utils import result_filter
+from idunn.instant_answer import normalize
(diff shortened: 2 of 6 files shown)


idunn/api/search.py
```diff
@@ -0,0 +1,115 @@
...
+    # Direct geocoding query
+    bragi_response = await bragi_client.raw_autocomplete(
+        {"q": query.q, "lang": query.lang, "limit": 5}
+    )
+    geocodings = (feature["properties"]["geocoding"] for feature in bragi_response["features"])
+    geocodings = sorted(
+        (
+            (result_filter.rank_bragi_response(query.q, feature), feature)
+            for feature in geocodings
+            if result_filter.check_bragi_response(query.q, feature)
```
amatissart, line 105: As this search is explicitly about Maps, we should probably be more permissive than for Instant Answers. For example, some extra tokens (that do not match the result) should probably be permitted.   
But this could probably be discussed in a separate PR.

idunn/api/search.py
```diff
@@ -0,0 +1,78 @@
...
+
+def search_single_place(place_id: str, lang: str):
+    try:
+        place = place_from_id(place_id, follow_redirect=True)
+    except Exception:
+        logger.warning("search: Failed to get place with id '%s'", place_id, exc_info=True)
+        return no_search_result()
+
+    detailed_place = place.load_place(lang=lang)
+    return IdunnAutocomplete(features=[detailed_place])
```
amatissart, line 39: this function is no longer used

idunn/api/instant_answer.py
```diff
@@ -181,7 +181,7 @@ async def get_instant_answer_intention(intention, query: str, lang: str):
             maps_frame_url=maps_urls.get_places_url(intention.filter, no_ui=True),
         )
 
-    return build_response(result, query=query, lang=lang)
+    return result
```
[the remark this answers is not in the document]
amatissart: This change still breaks the instant answer in some cases. For example on `?q=restaurant`, this function will return a 204 `Response`, and `build_response()` will fail.
It would probably be clearer to annotate `get_instant_answer_single_place` and `get_instant_answer_intention` to show that the return type is always `Response`. 
Anyway, it's no longer related to this PR, so maybe this change should be reverted for now, and be handled separately.

idunn/api/search.py
```diff
@@ -0,0 +1,66 @@
...
+                "request": {
+                    "query": query,
+                    "lang": lang,
+                }
+            },
+        )
+    return IdunnAutocomplete()
+
+
+async def search(query: QueryParams = Depends(QueryParams)) -> IdunnAutocomplete:
```
amatissart, line 30: Only `q` and `lang` query are supported. So the documentation would be misleading for this route.  For example `?nlu=false` would not have any effect.

Is there any obstacle to handle the same query parameters as `/v1/autocomplete` ?

remi-dupre: I agree with you that it sounds nicer to support the same set of parameters as `/v1/autocomplete`, I had some issues with it on a previous implementation but there is nothing holding it back now, this is fixed!

idunn/api/urls.py
```diff
@@ -74,6 +75,11 @@ def get_api_urls(settings):
             response_model=IdunnAutocomplete,
             response_model_exclude_unset=True,
         ),
+        APIRoute(
+            "/search",
+            search,
+            response_model=IdunnAutocomplete,
+        ),
```
amatissart, line 82: For consistency with "/autocomplete", `methods` and `response_model_exclude_unset` should be defined too.

idunn/api/search.py
```diff
@@ -0,0 +1,65 @@
...
+) -> IdunnAutocomplete:
+    """
+    Perform a query which is intended to display a relevant result directly (as
+    opposed to `autocomplete` which gives a list of plausible results).
+
+    Similarly to `instant_answer`, the result will need some quality checks.
+    """
+    query.q = normalize(query.q)
+
+    if query.lang in nlu_allowed_languages:
```
amatissart, line 41:
```suggestion
    if query.nlu and query.lang in nlu_allowed_languages:
```

remi-dupre: We also want to keep this option overridden by [AUTOCOMPLETE_NLU_SHADOW_ENABLED](https://github.com/Qwant/idunn/blob/master/idunn/utils/default_settings.yaml#L130), right?

amatissart: Well, this option has never been actually used, and is probably not very useful now that the `nlu=true` is used by the application in most cases.  
So I think it can be ignored for `/search`.

amatissart (approved):

PR #225, opened 2021-04-26
directions: add start/end time to direction outputs

remi-dupre:
Add start/end times to direction formats:

- for public transports, it is read directly from combigo API
- for other kind of transports, we assume that the travel should starts immediately and deduce the end time from the duration

This is outputed as an UTC iso string. An alternative would be to output local times (local in respect to start/end positions), but I don't think it would be a could idea, consider for example the corner case where a route would cross a timezone border which could be very missleading.


tests/test_directions.py
```diff
@@ -72,6 +74,8 @@ def test_direction_car(mock_directions_car):
     response_data = response.json()
     assert response_data["status"] == "success"
     assert len(response_data["data"]["routes"]) == 3
+    assert response_data["data"]["routes"][0]["start_time"] == "2018-06-14T08:30:00"
+    assert response_data["data"]["routes"][0]["end_time"] == "2018-06-14T09:00:19.200000"
```
amatissart, line 78: I would suggest to use `dt.isoformat(timespec="seconds")` to return a datetime with fixed precision.

idunn/directions/models.py
```diff
@@ -198,6 +201,18 @@ def __init__(self, **data):
...
             data["geometry"] = {"type": "FeatureCollection", "features": features_list}
 
+        if "start_time" not in data or "end_time" not in data:
+            if "dTime" in data and "aTime" in data:
+                # Handle combigo's output format
+                start = datetime.utcfromtimestamp(int(data["dTime"]) / 1000)
+                end = datetime.utcfromtimestamp(int(data["aTime"]) / 1000)
+            else:
+                start = datetime.utcnow()
+                end = start + timedelta(0, data.get("duration", 0))
```
amatissart, line 211: Nitpick: 
```suggestion
                end = start + timedelta(seconds=data.get("duration", 0))
```
looks clearer to me.

amatissart (commented):
I wonder whether returning UTC could be impractical here.  
If the application is expected to show local times, then the API should probably be responsible for finding the appropriate timezones for each endpoints, and converting timestamps.

amatissart (approved):

PR #226, opened 2021-04-29
`search`: more permissive filters

remi-dupre:
Allow the `search` endpoint to be a bit more permissive than `instant_answers`:

 - input query can contain only prefix of the result words (without spelling mistakes)
 - part of the input query is allowed not to match (at least 3 words must be relevant)


amatissart:
It looks good to me :+1:   
I wonder whether the Levenshtein distance could be set to a value > 1 for "search" results, but this kind of adjustments can be discussed later.

There is a conflict to be solved, #224 will probably be merged before anyway.

amatissart (approved):

PR #233, opened 2021-05-07
result_filter: consider all whitespaces as word separators

amatissart:
Match all Unicode whitespace characters as word separators, using [`\s`](https://docs.python.org/3.8/library/re.html#index-30).


remi-dupre (commented):
:+1:

remi-dupre (approved):

PR #235, opened 2021-05-17
instant_answer: direct queries to pages jaunes

remi-dupre:
Query pages jaunes for all queries to instant answers (which are not tagged as addresses), if the geocoder has no result this will be used as a fallback.

This also modifies the `nlu_client` which will now have a field for the "type" of intention to help understanding the nature of the request.

### DIFF
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -236,7 +236,6 @@ async def get_instant_answer(
                     IntentionType.BRAND,
                     IntentionType.CATEGORY,
                     IntentionType.POI,
-                    IntentionType.ANY_PLACE,
                 ],
             )
             if intentions and intentions[0].type in [
@@ -258,7 +257,7 @@ async def get_instant_answer(
         return await run_in_threadpool(
             pj_source.search_places,
             normalized_query,
-            intentions[0]._place_in_query,
+            intentions[0].description._place_in_query,
         )
 
     bragi_response, pj_response = await asyncio.gather(
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -1,6 +1,6 @@
 import asyncio
 import logging
-from fastapi import HTTPException, Query
+from fastapi import Query
 from fastapi.responses import ORJSONResponse, Response
 from fastapi.concurrency import run_in_threadpool
 from typing import Optional, List, Tuple
@@ -203,7 +203,9 @@ async def get_instant_answer(
     run more restrictive checks on its results.
     """
     normalized_query = normalize(q)
-    user_country = user_country.lower()
+
+    if user_country is not None:
+        user_country = user_country.lower()
 
     if len(normalized_query) > ia_max_query_length:
         return no_instant_answer(query=q, lang=lang, region=user_country)
@@ -253,8 +255,11 @@ async def get_instant_answer(
         if not (settings["IA_CALL_PJ_POI"] and user_country == "fr" and intentions):
             return []
 
-        place_in_query = any(it.description.place for it in intentions)
-        return await run_in_threadpool(pj_source.search_places, normalized_query, place_in_query)
+        return await run_in_threadpool(
+            pj_source.search_places,
+            normalized_query,
+            intentions.place_in_query,
+        )
 
     bragi_response, pj_response = await asyncio.gather(
         bragi_client.autocomplete(query), fetch_pj_response()
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -203,6 +203,7 @@ async def get_instant_answer(
     run more restrictive checks on its results.
     """
     normalized_query = normalize(q)
+    user_country = user_country.lower()
 
     if len(normalized_query) > ia_max_query_length:
         return no_instant_answer(query=q, lang=lang, region=user_country)
@@ -253,16 +254,7 @@ async def get_instant_answer(
             return []
 
         place_in_query = any(it.description.place for it in intentions)
-
-        try:
-            return await run_in_threadpool(
-                pj_source.search_places,
-                normalized_query,
-                place_in_query,
-            )
-        except HTTPException as exc:
-            logger.warning("Failed to query pagejaunes: %s", exc)
-            return []
+        return await run_in_threadpool(pj_source.search_places, normalized_query, place_in_query)
 
     bragi_response, pj_response = await asyncio.gather(
         bragi_client.autocomplete(query), fetch_pj_response()
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 from fastapi import Query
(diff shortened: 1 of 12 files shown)


idunn/geocoder/nlu_client.py
```diff
@@ -275,7 +305,13 @@ async def post_intentions(self, text, lang):
         response_nlu.raise_for_status()
         return response_nlu
 
-    async def get_intentions(self, text, lang, extra_geocoder_params=None) -> [Intention]:
+    async def get_intentions(
+        self,
+        text,
+        lang,
+        extra_geocoder_params=None,
+        allow_types=[IntentionType.BRAND, IntentionType.BRAND],
```
amatissart, line 313:
```suggestion
        allow_types=[IntentionType.BRAND, IntentionType.CATEGORY],
```

idunn/geocoder/nlu_client.py
```diff
@@ -212,27 +226,43 @@ async def get_category():
...
 
     @classmethod
     def is_poi_request(cls, tags_list):
         """Check if a request is addressed to a POI"""
         return any(t.get("tag") in NLU_POI_TAGS for t in tags_list)
 
+    @classmethod
+    def is_street_request(cls, tags_list):
+        """Check if a request is addressed to a POI"""
+        return all(t.get("tag") in NLU_STREET_TAGS for t in tags_list)
```
amatissart, line 264: What is the expected behavior here?  
On one hand, if any query related to a street or an address should match, then I think that other tags such as "city" should be allowed too.
On the other hand, similarly to `is_poi_request` we could also check that any "street" tag is present, using `any` instead of `all` to catch all queries where an address part could disturb a direct search.

idunn/geocoder/models/geocodejson.py
```diff
@@ -151,6 +152,14 @@ class Feature(BaseModel):
     context: Optional[Context]
 
 
+class IntentionType(Enum):
+    BRAND = "brand"
+    CATEGORY = "category"
+    PLACE = "place"
+    POI = "poi"
+    STREET = "street"
```
amatissart, line 160: As these values are theoretically exposed by the API, I think that we need to be particularly careful about the naming and restrain unwanted dependency with the current tagger implementation. 
For example, 
* `"street"` would also include intention about exact addresses, so `"address"` could be clearer ?
* `"place"` sounds confusing to me: would `"locality"` be correct ? or is it a generic fallback case, and then `"any_place" ` could be more explicit ?

idunn/geocoder/models/geocodejson.py
```diff
@@ -165,6 +174,7 @@ class IntentionDescription(BaseModel):
 
 
 class Intention(BaseModel):
+    type: IntentionType
```
amatissart, line 177: There are also many other "type" fields in the API and in the implementation,
but I don't have any better name right now :slightly_smiling_face:

remi-dupre: We could go for `mode`, `kind`, ... but I can't come up with something that doesn't just look like a synonym :(

idunn/geocoder/nlu_client.py
```diff
@@ -212,27 +226,57 @@ async def get_category():
...
+        True
+        >>> NLU_Helper.is_street_request([
+        ...     {"phrase": "restaurant", "tag": "cat"},
+        ...     {"phrase": "rue des Rosiers, Paris", "tag": "street"},
+        ... ])
+        False
+        """
+        contains_street_tag = any(t.get("tag") in NLU_STREET_TAGS for t in tags_list)
+        only_place_tags = all(t.get("tag") in NLU_STREET_TAGS + NLU_PLACE_TAGS for t in tags_list)
+        return contains_street_tag and only_place_tags
```
amatissart, line 278: I am still a bit reluctant to this implementation where any query like "Paris" (single tag `city`) will trigger an additional request to PJ, with little benefit.  But this can probably be discussed/adjusted separately.

idunn/api/instant_answer.py
```diff
@@ -223,34 +229,62 @@ async def get_instant_answer(
...
-        key=lambda item: -item[0],  # sort by descending rank
+
+    async def fetch_pj_response():
+        if not (settings["IA_CALL_PJ_POI"] and user_country == "fr" and intentions):
+            return []
+
+        return await run_in_threadpool(
+            pj_source.search_places,
+            normalized_query,
+            intentions.place_in_query,
```
amatissart, line 261: `intentions` is a list, right?  
```suggestion
            intentions[0].place_in_query,
```

idunn/geocoder/models/geocodejson.py
```diff
@@ -171,6 +171,7 @@ class IntentionDescription(BaseModel):
     query: Optional[str]
     category: Optional[str]
     place: Optional[Feature]
+    place_in_query: bool = False
```
amatissart, line 174: I'm not very fond of including this implementation detail in the API.  
Perhaps [`PrivateAttr`](https://pydantic-docs.helpmanual.io/usage/models/#private-model-attributes) could be used instead, although it would need to be set explicitly, and not via `__init__`.

remi-dupre: This sounds to fit perfectly the use case, maybe it could also be used for `type` ?

amatissart: As you wish. `type` looks well-defined enough. I don't mind keeping it that way.   
And using private attributes can be cumbersome as they can't be set by the default constructor (as far as I can see).

idunn/api/instant_answer.py
```diff
@@ -223,34 +229,62 @@ async def get_instant_answer(
...
-        key=lambda item: -item[0],  # sort by descending rank
+
+    async def fetch_pj_response():
+        if not (settings["IA_CALL_PJ_POI"] and user_country == "fr" and intentions):
+            return []
+
+        return await run_in_threadpool(
+            pj_source.search_places,
+            normalized_query,
+            intentions[0]._place_in_query,
```
amatissart, line 261:
```suggestion
            intentions[0].description._place_in_query,
```

Do you think a test could be added to cover this case?

idunn/api/instant_answer.py
```diff
@@ -223,34 +229,62 @@ async def get_instant_answer(
...
             intentions = await nlu_client.get_intentions(
-                normalized_query, lang, extra_geocoder_params
+                normalized_query,
+                lang,
+                extra_geocoder_params,
+                allow_types=[
+                    IntentionType.BRAND,
+                    IntentionType.CATEGORY,
+                    IntentionType.POI,
+                    IntentionType.ANY_PLACE,
```
amatissart, line 239: As this intention type `ANY_PLACE` is very common and could have a significant impact on response times, I wonder if we could remove it from this list for now, and hopefully improve the coverage of intentions with future iterations of the tagger?

@remi-dupre  What do you think ?

remi-dupre: I'll change it, especially since this a rather niche improvement :/

amatissart (approved):

PR #238, opened 2021-05-20
result_filter: Allow match on postcode only, and on postcode prefix for longer queries

amatissart:
The expected behavior is described in additional test cases.  

* a query that consists of a postcode only should match any result with that postcode.
* a postcode prefix (with at least 2 characters) should be considered as a matching word in the query.  
This is especially useful to accept "départements" numbers in France.

### DIFF
--- a/idunn/utils/result_filter.py
+++ b/idunn/utils/result_filter.py
@@ -238,11 +238,11 @@ class ResultFilter:
         candidate_terms_to_cover = (
             terms
             for name_words in map(self.words, names)
-            for terms in [
-                name_words,
-                *(name_words + self.words(admin) for admin in admins),
-                *(name_words + [postcode] for postcode in postcodes),
-            ]
+            for terms in chain(
+                [name_words],
+                (name_words + self.words(admin) for admin in admins),
+                (name_words + [postcode] for postcode in postcodes),
+            )
         )
         if len(query_words) == 1:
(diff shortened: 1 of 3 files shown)


idunn/utils/result_filter.py
```diff
@@ -221,21 +228,28 @@ def check(
...
             for terms in [
                 name_words,
                 *(name_words + self.words(admin) for admin in admins),
                 *(name_words + [postcode] for postcode in postcodes),
             ]
         )
+        if len(query_words) == 1:
+            candidate_terms_to_cover = chain(
+                candidate_terms_to_cover, ([postcode] for postcode in postcodes)
+            )
```
remi-dupre, line 250: I wonder if it would make it slightly more readable to use `candidate_terms_to_cover` as a list here? This is definitely opinionated and not important. I just, as usual, enjoy this kind of reflections ^^'

remi-dupre (dismissed):

idunn/utils/result_filter.py
amatissart: In most cases, it's probably insignificant. But in some extreme cases, (with many names variants :heavy_multiplication_x:  many admins), there may be a real benefit to evaluate `candidate_terms_to_cover` lazily. So I tried to preserve this behavior.

PR #246, opened 2021-06-03
add NLU_CLIENT_TIMEOUT setting

remi-dupre:
When using a slow or unstable network the hard-coded 300ms timeout in `nlu_client` can be rather short, this is just a tiny PR to suggest making it configurable.


amatissart:
@remi-dupre  Looks good to me. But this client is not specific to the /autocomplete. What about `NLU_CLIENT_TIMEOUT` ?

amatissart (approved):

PR #249, opened 2021-06-11
Extend "raw_filters" in categories definitions for OSM POIs, to allow "cuisine" filter

amatissart:
In "categories.yaml", the `raw_filters` (used in ES queries) are now defined with objects that accept the following keys : `class`, `subclass`, `cuisine`.   
The new `MimirPoiFilter` is responsible for translating these values into appropriate `poi_type.name` values (see https://github.com/Qwant/fafnir/pull/61)

All functions related to queries in ES for documents built by Mimirsbrunn importers are moved to "idunn/datasources/mimirsbrunn.py".

The new implementation removes a deprecated behavior about "class" filters: both `{poi_class}` and `class_{poi_class}` were accepted. The first format is obsolete. That's why a few existing fixtures need to be updated.

### DIFF
--- a/idunn/utils/categories.yml
+++ b/idunn/utils/categories.yml
@@ -91,7 +91,7 @@ categories:
           - class: "restaurant"
             cuisine: "burger"
           - class: "fast_food"
-            cuisine: "pizza"
+            cuisine: "burger"
         pj_what: "burger"
 
     food_italian:
--- a/idunn/utils/categories.yml
+++ b/idunn/utils/categories.yml
@@ -6,13 +6,12 @@ categories:
         pj_filters:
           - "salles de cinéma"
         raw_filters:
-          - "cinema,*"
+          - class: "cinema"
 
     # NOTE: This is a legacy alias of `health_hospital`.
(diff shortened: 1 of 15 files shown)


idunn/utils/categories.yml
```diff
@@ -70,514 +69,563 @@ categories:
...
+            cuisine: "pizza"
         pj_what: "pizzeria"
 
     food_burger:
         raw_filters:
-          - "restaurant,burger"
+          - class: "restaurant"
+            cuisine: "burger"
+          - class: "fast_food"
+            cuisine: "pizza"
```
remi-dupre, line 94: Seems like a small typo:

```suggestion
            cuisine: "burger"
```

remi-dupre (dismissed):
I feel like those filters are even more readable in this way :+1:

idunn/utils/categories.yml
amatissart: Thanks! It was certain that I would miss something in this file.

remi-dupre (approved):

PR #250, opened 2021-06-14
pj: new "transactional" block

remi-dupre:
Add a block "transactional" which adds new links for ordering/booking/taking an appointment.

Links from `RESERVER_LA_FOURCHETTE` seem broken which doesn't seem critical as it usually backed up with other values, but we should probably still query PJ about it.


remi-dupre:
Do you think we should drop support for the "legacy" PJ source now that we appear to have a stable use of the official API and started adding specific features?

idunn/blocks/commercial.py
```diff
@@ -0,0 +1,24 @@
...
+        args = {
+            "booking_url": place.get_booking_url(),
+            "appointment_url": place.get_appointment_url(),
+            "order_url": place.get_appointment_url(),
+        }
+
+        args = {key: val for key, val in args.items() if val}
+
+        if not args:
+            return None
```
amatissart, line 22: Is this transformation only useful to check that a non-empty value exists?  
If so, it could be more explicit to write:
```suggestion
        if not any(args.values()):
            return None
```

idunn/blocks/commercial.py
```diff
@@ -0,0 +1,24 @@
+from .base import BaseBlock
+from typing import Literal, Optional
+
+
+class CommercialBlock(BaseBlock):
+    type: Literal["commercial"] = "commercial"
```
amatissart, line 6: I am a bit uncertain about this block name. An appointment may be relevant for non-commercial POIs (administration, health, etc.). Would `transactional` be better ?

idunn/places/pj_poi.py
```diff
@@ -419,3 +420,48 @@ def get_raw_grades(self):
 
     def get_reviews_url(self):
         return self.get_source_url() + "#ancreBlocAvis"
+
+    def get_booking_url(self):
+        return next(
+            (
+                link.url
```
amatissart, line 427: These new functions should use `resolve_url()` to build a URL that will be resolved on server-side, similarly to `get_website()`.

idunn/blocks/commercial.py
```diff
@@ -0,0 +1,24 @@
...
+    booking_url: Optional[str]
+    appointment_url: Optional[str]
+    order_url: Optional[str]
```
amatissart, line 9: An additional field `quotation_request_url` could be defined using the transactional link with type `QUOTATION_REQUEST`.

idunn/blocks/commercial.py
remi-dupre: :+1: 
I was uncomfortable with all the `none` fields this block could add an hoped that [`response_model_exclude_unset`](https://fastapi.tiangolo.com/tutorial/response-model/#use-the-response_model_exclude_unset-parameter) would help me here, but it's not even designed to solve this problem...

idunn/places/models/pj_info.py
```diff
@@ -48,6 +52,33 @@ class ContactInfo(BaseModel):
...
 
 
+class TransactionalLinkType(Enum):
+    RESERVER = "RESERVER"
+    RESERVER_INTERNE = "RESERVER_INTERNE"
+    COMMANDER = "COMMANDER"
+    COMMANDER_CHRONO = "COMMANDER_CHRONO"
+    QUOTATION_REQUEST = "QUOTATION_REQUEST"
+    PRENDRE_RDV_EXTERNE = "PRENDRE_RDV_EXTERNE"
+    PRENDRE_RDV_INTERNE = "PRENDRE_RDV_INTERNE"
```
amatissart, line 62: one more possibility:
```python
    PRENDRE_RDV_CLIC_RDV = "PRENDRE_RDV_CLIC_RDV"
```

amatissart (approved):

PR #252, opened 2021-06-17
categories.yml: update definitions for leisure and administrative (pj)

remi-dupre:
Leisure contains most POIs from other categories cinema, theatre and attraction. From attraction I didn't included POIs with sublass `attraction` or `artwork` since most of them are historic places (I think leisure should be related to activities).

Screenshot_2021-06-17_08-15-23

For admistrative POIs, the search "service public" seems slightly off-topic but gives quite convincing results:

| Screenshot_2021-06-17_08-30-54 | Screenshot_2021-06-17_08-30-32 |
|---|---|


idunn/utils/categories.yml
```diff
@@ -1,12 +1,22 @@
...
+          - class: "theatre"
+          - subclass: "nightclub"
+          - subclass: "escape_game"
+          - subclass: "golf_course"
+          - subclass: "miniature_golf"
+          - class: "aquarium"
+            subclass: "aquarium"
+          - subclass: "art_gallery"
+          - subclass: "art_center"
+          - subclass: "gallery"
```
amatissart, line 16: `art_gallery` does not seem to be a valid subclass. And `art_center` is actually `arts_centre`.  
(Yes this nomenclature is a bit of a mess).
```suggestion
          - subclass: "arts_center"
          - subclass: "gallery"
```

amatissart (approved):

PR #258, opened 2021-07-01
Add new block: social

remi-dupre:
This adds new links for facebook, twitter, instagram and youtube (this last one is not really a social network but it has consequent coverage on OSM). Additionally, Facebook links won't be used anymore as websites urls (previously as a fallback for OSM and depending on the order for PJ).

~~I couldn't find any example of PJ places with an Instagram link so I didn't include it here :/~~

### DIFF
--- a/idunn/places/pj_poi.py
+++ b/idunn/places/pj_poi.py
@@ -284,14 +284,14 @@ class PjApiPOI(BasePlace):
         )
 
     def get_website(self):
-        return self.get_website_url_for_type(UrlType.NON_SOCIAL)
+        return self.get_website_url_for_type(UrlType.EXTERNAL_WEBSITE)
 
     def get_website_label(self):
         if isinstance(self.data, pj_find.Listing):
             # FIXME: Ideally the Listing would include a "suggested_label" too
             return self.get_local_name()
 
-        suggested_label = self.get_website_url_for_type(UrlType.NON_SOCIAL)
+        suggested_label = self.get_website_url_for_type(UrlType.EXTERNAL_WEBSITE)
         prefix = "Voir le site "
 
         if not suggested_label:
--- a/idunn/places/pj_poi.py
+++ b/idunn/places/pj_poi.py
@@ -291,8 +291,15 @@ class PjApiPOI(BasePlace):
             # FIXME: Ideally the Listing would include a "suggested_label" too
             return self.get_local_name()
 
-        suggested_label = self.get_website_url_for_type(UrlType.EXTERNAL_WEBSITE)
         prefix = "Voir le site "
+        suggested_label = next(
+            (
+                website.suggested_label
+                for website in self.data.website_urls or []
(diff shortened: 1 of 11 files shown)


tests/fixtures/orsay_museum.json
```diff
@@ -277,6 +277,26 @@
...
+    },
+    {
+        "key": "contact:twitter",
+        "value": "@MuseeOrsay"
+    },
+    {
+        "key": "contact:instagram",
+        "value": "https://www.instagram.com/MuseeOrsay"
+    },
+    {
```
bbecquet, line 292: This key is duplicated. Is it on purpose?

idunn/places/pj_poi.py
```diff
@@ -273,28 +273,47 @@ def get_phone(self):
...
     def get_website_label(self):
-        if not self.data.website_urls:
-            return None
-
         if isinstance(self.data, pj_find.Listing):
             # FIXME: Ideally the Listing would include a "suggested_label" too
             return self.get_local_name()
 
-        suggested_label = self.data.website_urls[0].suggested_label
+        suggested_label = self.get_website_url_for_type(UrlType.EXTERNAL_WEBSITE)
```
amatissart, line 294: So the `suggested_label` field provided by the API is no longer used ?

tests/fixtures/orsay_museum.json
remi-dupre: Nice catch :facepalm:

idunn/places/pj_poi.py
remi-dupre: That's fixed, and also the lack of test :sweat_smile:

idunn/places/pj_poi.py
```diff
@@ -273,28 +273,54 @@ def get_phone(self):
             None,
         )
 
-    def get_website(self):
-        if not self.data.website_urls:
-            return None
+    def get_website_url_for_type(self, social_network: UrlType) -> Optional[str]:
```
bbecquet, line 276: Maybe this could be turned into a `get_website_for_type` method so it can be re-used for url and label?

remi-dupre: It could, but I think it would move the pain to applying `resolve_url` to the results only when it is not `None`, which would require an extra branching after most calls:

```python
    def get_website(self):
        website = self.get_website_url_for_type(UrlType.EXTERNAL_WEBSITE)

        if not website:
            return None

        return resolve_url(website.website_url)
```

Python's `None` is actually quite painful to work with as there is not much syntactic sugar for common operations (such as optional chaining on JS or the awesome set of [utilities](https://doc.rust-lang.org/std/option/enum.Option.html) and `?` operator in Rust) :cry:

bbecquet: I get your point :+1:

bbecquet (approved):

amatissart (approved):

PR #261, opened 2021-07-15
Add source `wiki_es`

remi-dupre:
Reorganize calls to the Wikipedia API to move most logic in `datasources/`, this was also an opportunity to de-duplicate the calls to Redis cache.

### DIFF
--- a/idunn/datasources/wikidata.py
+++ b/idunn/datasources/wikidata.py
@@ -1,0 +1,98 @@
+from elasticsearch import (
+    Elasticsearch,
+    ConnectionError,
+    NotFoundError,
+    ElasticsearchException,
+)
+import logging
+from idunn import settings
+from idunn.utils import prometheus
+from idunn.utils.redis import RedisWrapper
+
+
+logger = logging.getLogger(__name__)
+
+
+class WikidataEs:
+    REDIS_INFO_KEY_PREFIX = "get_wiki_info"
+
+    def __init__(self):
+        wiki_es_url = settings.get("WIKI_ES")
+        wiki_es_max_retries = settings.get("WIKI_ES_MAX_RETRIES")
+        wiki_es_timeout = settings.get("WIKI_ES_TIMEOUT")
+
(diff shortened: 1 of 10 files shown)


idunn/places/base.py
```diff
@@ -34,29 +33,20 @@ def __init__(self, d):
...
         self.properties = {}
 
-    def get_wiki_info(self, wikidata_id, wiki_index):
-        return WikidataConnector.get_wiki_info(wikidata_id, wiki_index)
+    def get_wiki_info(self, wikidata_id, lang):
+        return wikidata_es.get_info(wikidata_id, lang)
 
     def get_wiki_index(self, lang):
-        return WikidataConnector.get_wiki_index(lang)
+        return wikidata_es.get_index(lang)
```
amatissart, line 40: This two methods `get_wiki_info` and ` get_wiki_index` could probably be removed completely.  
`get_wiki_resp` (or any better name) should be the only entrypoint to fetch the Wikipedia data associated to a POI.

idunn/datasources/wikidata.py
```diff
@@ -0,0 +1,98 @@
...
+import logging
+from idunn import settings
+from idunn.utils import prometheus
+from idunn.utils.redis import RedisWrapper
+
+
+logger = logging.getLogger(__name__)
+
+
+class WikidataEs:
```
amatissart, line 16: This name is consistent with the names of the ES indices (`wikidata_en`,  `wikidata_fr`, etc.)   
But it's still a bit confusing, as the data is actually about Wikipedia articles.  
Maybe a comment would be useful ?

remi-dupre: I'll add a short description of this and also rename it to `WikiEs`, it doesn't really fix the confusion but at least doesn't make it worst. As I understood these indexes may be totally unrelated to Wikidata as there was just a confusion to what Wikidata actually was when it was built?

amatissart (approved):

PR #263, opened 2021-07-20
add block `description`

remi-dupre:
Adds a new `description` block, which contains:

 - the description itself
 - the source of the description (which would be helpful to display labels like "read more on Wikipedia")
 - optionally, the url to the source

The information may come from several sources:

 - Wikipedia if available (which means that the old `wikipedia` block is now deprecated)
 - `description:lang` for OSM. Should we fallback to `decription` without a lang ? It would give much more data but at the cost of loosing some consistency with language, I'm not sure what users actually expect ? (In most case they wouldn't notice as you typically use QMaps in the country you currently live in)
 - `description` for pagesjaunes, although it is only rarely available : we should ask about cases were a description is visible on their website but not in the API :confused: 

For example for [this place](http://localhost:3000/place/pj:56110744@Brasserie_Bofinger#map=16.50/48.8538800/2.3678450) the block is:

```json
{
    "type": "description",
    "description": "DINER SOUS LA COUPOLE - Situé entre Bastille et la place des Vosges, le Bofinger vous reçoit dans un cadre d'exception emprunt d'histoire et de mémoire. Cette élégante brasserie alsacienne du 19e siècle saura vous surprendre avec sa coupole en verre qui illumine la grande salle, ses toiles du célèbre Hansi, ses boiseries, céramiques et miroirs d'époque... POUR LES GOURMETS - Du côté des papilles, le Chef Georges Belondrade met tout en œuvre pour éveiller vos sens. Il travaille poissons et fruits de mer tout en respectant la tradition Bofinger, et vous propose une carte variée qui ravira les fins gourmets. En dessert, vous craquerez volontiers pour le kougelhopf glacé aux griottines sauvages... A NE PAS MANQUER - Une adresse parisienne incontournable qui vous transportera au début des Années Folles. A découvrir sans plus tarder !",
    "source": "pagesjaunes",
    "url": null
}
```

### DIFF
--- a/idunn/blocks/description.py
+++ b/idunn/blocks/description.py
@@ -1,0 +1,48 @@
+from enum import Enum
+from typing import Literal, Optional
+
+import idunn
+from idunn import settings
+from idunn.blocks.wikipedia import WikipediaBlock
+from .base import BaseBlock
+
+DESC_MAX_SIZE = int(settings["DESC_MAX_SIZE"])
+
+
+def limit_size(content: str) -> str:
+    if len(content) > DESC_MAX_SIZE:
+        return content[:DESC_MAX_SIZE] + "…"
+
+    return content
+
+
+class DescriptionSources(Enum):
+    OSM = "osm"
+    PAGESJAUNES = "pagesjaunes"
+    WIKIPEDIA = "wikipedia"
+
+
+class DescriptionBlock(BaseBlock):
+    type: Literal["description"] = "description"
+    description: str
+    source: DescriptionSources
+    url: Optional[str]
+
+    @classmethod
+    def from_es(cls, place, lang):
+        if wiki_block := WikipediaBlock.from_es(place, lang):
+            return cls(
+                description=wiki_block.description,
+                source=DescriptionSources.WIKIPEDIA,
+                url=wiki_block.url,
+            )
+
+        if description := place.get_description(lang):
+            if isinstance(place, idunn.places.PjApiPOI):
+                source = DescriptionSources.PAGESJAUNES
+            else:
+                source = DescriptionSources.OSM
+
+            return cls(description=description, source=source)
+
+        return None
--- a/idunn/blocks/description.py
+++ b/idunn/blocks/description.py
@@ -43,6 +43,6 @@ class DescriptionBlock(BaseBlock):
             else:
                 source = DescriptionSources.OSM
 
-            return cls(description=description, source=source)
+            return cls(description=description, source=source, url=place.get_description_url(lang))
 
         return None
--- a/idunn/blocks/description.py
+++ b/idunn/blocks/description.py
@@ -32,7 +32,7 @@ class DescriptionBlock(BaseBlock):
     def from_es(cls, place, lang):
         if wiki_block := WikipediaBlock.from_es(place, lang):
             return cls(
-                description=wiki_block.description,
+                description=limit_size(wiki_block.description),
                 source=DescriptionSources.WIKIPEDIA,
                 url=wiki_block.url,
             )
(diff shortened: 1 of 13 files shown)


idunn/places/pj_poi.py
```diff
@@ -506,3 +506,15 @@ def get_order_url(self):
 
     def get_quotation_request_url(self):
         return self.get_transactional_url([TransactionalLinkType.QUOTATION_REQUEST])
+
+    def get_description(self, lang):
+        if lang != "fr" or isinstance(self.data, pj_find.Listing):
```
bbecquet, line 511: Sorry, maybe a dumb question, but what does this `pj_find.Listing` mean?

tests/test_api_circuit_breaker.py
```diff
@@ -75,7 +75,7 @@ def test_circuit_breaker_404(breaker_test, disable_redis):
             response = client.get(url="http://localhost/v1/pois/osm:way:7777777?lang=es")
 
         assert breaker_test.current_state == "closed"
-        assert len(rsps.calls) == 4
+        assert len(rsps.calls) == 8
```
bbecquet, line 78: Why do we need these changes of number of calls, here and in `test_rate_limiter.py` and `test_wiki_ES.py`?

remi-dupre: Now the API returns two blocks with content with Wikipedia API which doubles the calls to the API. In practice Idunn uses an internal cache to avoid those extra calls, but it is disabled for this tests as the circuit breaker (which is being tested) acts behind this cache.

This is a bit messy and there may be a way to refactor things a little (to rely less on the cache), but I think we could just not bother too much about it and just remove the Wikipedia block as soon as this new block is used by erdapfel in production.

idunn/places/pj_poi.py
remi-dupre: This is the root model for data coming from pagejaune's endpoint "find", which is used to search for POIs. This endpoint is mostly similar to the other endpoint "info" which gives details about a POIs, but there are some subtle differences in there response format that make it hard to build a common representation.

Here, the description is just not expected to be found in the "find" endpoint, but we wouldn't display it anyway (except for some quite rare cases in the instant answer).

idunn/blocks/description.py
```diff
@@ -0,0 +1,48 @@
...
+from .base import BaseBlock
+
+DESC_MAX_SIZE = int(settings["DESC_MAX_SIZE"])
+
+
+def limit_size(content: str) -> str:
+    if len(content) > DESC_MAX_SIZE:
+        return content[:DESC_MAX_SIZE] + "…"
+
+    return content
```
amatissart, line 16: This function is not used, as far as I can see.

tests/test_rate_limiter.py
```diff
@@ -63,7 +63,7 @@ def test_rate_limiter_with_redis(limiter_test_normal, mock_wikipedia_response):
     Test that Idunn stops external requests when
     we are above the max rate
 
-    We mock 5 calls to wikipedia while the max number
+    We mock 5*2 calls to wikipedia while the max number
     of calls is 3*2, so we test that after 3 calls
     no more calls are done
     """
```
amatissart, line 69: This comment could probably be made clearer.  

5 requests to Idunn, and each of them requires 2 blocks ("description" + "wikipedia") * 2 requests ("title" + "summary") = 4 calls per POI request.  
The limit is set to 12 calls to Wikipedia, so after 3 POI requests, these blocks are no longer present.

idunn/blocks/description.py
remi-dupre: :man_facepalming: 

I'll just apply it to all description results, call it on all description outputs, which will be consistent with former Wikipedia data and extend the same rule to pagejaunes and OSM (even though it won't make a huge difference for these two as their description field is usually short).

amatissart (approved):

PR #264, opened 2021-07-21
add block `delivery`

remi-dupre:
Add a new `delivery` block which list 'delivery' options.

For example with [this poi](http://localhost:3000/place/pj:08692611@DARTY_Montparnasse) it outputs following block:

```json
{
    "type": "delivery",
    "click_and_collect": "yes",
    "takeaway": "unknown",
    "delivery": "yes"
}
```

EDIT: I finally added support for basic OSM tags for [`takeaway=yes`](https://taginfo.openstreetmap.org/keys/takeaway#values) and [`delivery=yes`](https://taginfo.openstreetmap.org/keys/?key=delivery#values), since it's basically free and stills ads information for respectively about 252k and 37k POIs.

### DIFF
--- a/idunn/blocks/delivery.py
+++ b/idunn/blocks/delivery.py
@@ -1,32 +1,33 @@
 from enum import Enum
-from typing import List, Literal
+from typing import Literal
 
 from .base import BaseBlock
 
 
-class DeliveryType(Enum):
-    CLICK_AND_COLLECT = "click_and_collect"
-    DELIVERY = "delivery"
-    TAKEAWAY = "takeaway"
+class DeliveryState(Enum):
+    YES = "yes"
+    UNKNOWN = "unknown"
+
+    @classmethod
+    def from_bool(cls, is_yes: bool):
+        return cls.YES if is_yes else cls.UNKNOWN
 
 
 class DeliveryBlock(BaseBlock):
     type: Literal["delivery"] = "delivery"
-    available: List[DeliveryType]
+    click_and_collect: DeliveryState
+    delivery: DeliveryState
+    takeaway: DeliveryState
 
     @classmethod
     def from_es(cls, place, lang):
-        available = [
-            delivery_type
-            for delivery_type, is_available in [
-                (DeliveryType.CLICK_AND_COLLECT, place.has_click_and_collect()),
-                (DeliveryType.DELIVERY, place.has_delivery()),
-                (DeliveryType.TAKEAWAY, place.has_takeaway()),
-            ]
-            if is_available
-        ]
+        states = {
+            "click_and_collect": DeliveryState.from_bool(place.has_click_and_collect()),
+            "delivery": DeliveryState.from_bool(place.has_delivery()),
+            "takeaway": DeliveryState.from_bool(place.has_takeaway()),
+        }
 
-        if not available:
+        if all(state == DeliveryState.UNKNOWN for state in states.values()):
             return None
 
-        return cls(available=available)
+        return cls(**states)
--- a/idunn/blocks/delivery.py
+++ b/idunn/blocks/delivery.py
@@ -1,0 +1,32 @@
+from enum import Enum
+from typing import List, Literal
+
+from .base import BaseBlock
+
+
+class DeliveryType(Enum):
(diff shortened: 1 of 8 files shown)


idunn/places/base.py
```diff
@@ -272,6 +272,15 @@ def get_appointment_url(self):
...
         return None
 
+    def has_click_and_collect(self):
+        return False
+
+    def has_delivery(self):
+        return self.properties.get("delivery") == "yes"
+
+    def has_takeaway(self):
+        return self.properties.get("takeaway") == "yes"
```
amatissart, line 282: Any reason not to include `takeaway=only` which has a significant usage according to https://taginfo.openstreetmap.org/keys/takeaway#values ?

```suggestion
    def has_takeaway(self):
        return self.properties.get("takeaway") in ("yes", "only")
```

tests/test_delivery.py
```diff
@@ -0,0 +1,10 @@
+from idunn.blocks.delivery import DeliveryBlock
+from idunn.places import POI
+
+
+def test_website_block():
+    web_block = DeliveryBlock.from_es(
+        POI({"properties": {"delivery": "yes", "takeaway": "yes"}}), lang="en"
+    )
+
+    assert web_block == DeliveryBlock(available=["delivery", "takeaway"])
```
amatissart, line 10: `test_website_block` and `web_block` should be renamed

idunn/places/pj_poi.py
```diff
@@ -9,15 +9,11 @@
...
-            "accès aux personnes à mobilité réduite",
-            "accès pour personne à mobilité réduite",
-        ]
-    )
-)
+
+CLICK_AND_COLLECT = re.compile(r"retrait .*")
+DELIVERY = re.compile(r"commande en ligne|livraison.*")
+TAKEAWAY = re.compile(r".* à emporter")
+WHEELCHAIN_ACCESSIBLE = re.compile("accès (handicapés?|(aux|pour) personnes? à mobilité réduite)")
```
amatissart, line 16: Let's take advantage of this change to fix the typo
```suggestion
WHEELCHAIR_ACCESSIBLE = re.compile("accès (handicapés?|(aux|pour) personnes? à mobilité réduite)")
```

idunn/blocks/delivery.py
```diff
@@ -0,0 +1,32 @@
...
+
+class DeliveryType(Enum):
+    CLICK_AND_COLLECT = "click_and_collect"
+    DELIVERY = "delivery"
+    TAKEAWAY = "takeaway"
+
+
+class DeliveryBlock(BaseBlock):
+    type: Literal["delivery"] = "delivery"
+    available: List[DeliveryType]
```
amatissart, line 15: Any specific reason to use a list here ?  

Explicit fields `click_and_collect`, `delivery`, `takeway` could be easier to maintain and more discoverable.   
Similarly to the `AccessibilityBlock`, the possible values would be `"yes"` or `"unknown"`.

tests/test_delivery.py
remi-dupre: As `web_block` is used in several other blocks I was wondering if it was meaning something ^^' `test_website_block` is a plain mistake though.

idunn/blocks/delivery.py
remi-dupre: I had no strong opinion about both options so I'll just trust yours (plus "we already do it the over way" seems like a good argument).

remi-dupre: Now it looks like that:

```json
{
    "type": "delivery",
    "click_and_collect": "yes",
    "takeaway": "unknown",
    "delivery": "yes"
}
```

tests/test_delivery.py
amatissart: I really think this is some kind of copy-paste contamination :smile:

amatissart:
@remi-dupre A rebase is required now that #263 has been merged.

amatissart:
It looks good to me.

@bbecquet Does the block format look good to you?

bbecquet:
> @bbecquet Does the block format look good to you?

Yes, ok for me too :+1:

bbecquet (approved):

PR #269, opened 2021-08-17
add new block `stars`

remi-dupre:
Add a new `stars` block which indicates if the POI is a starred restaurant or hotel and eventually gives its rating.

The format looks like follows:

```json
{
    "type": "stars",
    "ratings": [
        {"has_stars": "yes", "nb_stars": 4.0, "kind": "lodging"}
    ]
}
```

```json
{
    "type": "stars",
    "ratings": [
        {"has_stars": "yes", "nb_stars": null, "kind": "restaurant"}
    ]
}
```

### DIFF
--- a/idunn/blocks/stars.py
+++ b/idunn/blocks/stars.py
@@ -20,6 +20,13 @@ class StarsKind(Enum):
     LODGING = "lodging"
     RESTAURANT = "restaurant"
 
+    def check_class(self, class_name: Optional[str]) -> bool:
+        return (self, class_name) in [
+            (self.LODGING, "lodging"),
+            (self.RESTAURANT, "restaurant"),
+            (self.RESTAURANT, "fast_food"),
+        ]
+
 
 class StarsDetails(BaseModel):
     has_stars: StarsAvailable
@@ -47,7 +54,7 @@ class StarsBlock(BaseBlock):
                 (StarsKind.LODGING, place.get_lodging_stars()),
                 (StarsKind.RESTAURANT, place.get_restaurant_stars()),
             ]
-            if stars is not None
+            if stars is not None and kind.check_class(place.get_class_name())
         ]
 
         if not ratings:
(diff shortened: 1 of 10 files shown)


idunn/blocks/stars.py
```diff
@@ -0,0 +1,56 @@
...
+                    if isinstance(stars, bool)
+                    else StarsAvailable.YES
+                ),
+                nb_stars=stars if isinstance(stars, float) else None,
+                kind=kind,
+            )
+            for kind, stars in [
+                (StarsKind.LODGING, place.get_lodging_stars()),
+                (StarsKind.RESTAURANT, place.get_restaurant_stars()),
+            ]
```
amatissart, line 49: To help with the integration, perhaps we could keep only one of these ratings, depending on `place.get_class_name()` ?   
Then the client could just append the number of stars with the title of the poi class/subclass, without any redundant implementation on both sides, or risk of confusion.

cc @xem @bbecquet

remi-dupre: What I had in mind is that erdapfel would blindly display all items of the list, but maybe we risk displaying irrelevant data?

idunn/blocks/stars.py
```diff
@@ -20,6 +20,13 @@ class StarsKind(Enum):
     LODGING = "lodging"
     RESTAURANT = "restaurant"
 
+    def check_class(self, subclass: Optional[str]) -> bool:
+        return (self, subclass) in [
+            (self.LODGING, "lodging"),
+            (self.RESTAURANT, "restaurant"),
+            (self.RESTAURANT, "fast_food"),
+        ]
```
amatissart, line 28: `subclass` should rather be named `class_name`

remi-dupre: I just rewrote last commit as this is a very minor change.

amatissart (approved):

PR #272, opened 2021-09-22
pj_poi: Fix 'get_lodging_stars' on null category

amatissart:
```python
if match := self.HOTEL_STARS_REGEX.match(acc.category):
```
would fail if `acc.category` is `None`.

### DIFF
--- a/idunn/places/pj_poi.py
+++ b/idunn/places/pj_poi.py
@@ -555,12 +555,29 @@ class PjApiPOI(BasePlace):
     HOTEL_STARS_REGEX = re.compile(r"hôtel (?P<rating>\d) étoiles?")
 
     def get_lodging_stars(self) -> Optional[Union[bool, float]]:
+        """
+        >>> poi = PjApiPOI(pj_info.Response(**{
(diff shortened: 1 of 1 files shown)


idunn/places/pj_poi.py
```diff
@@ -555,12 +555,29 @@ def has_takeaway(self):
...
+        """
         if isinstance(self.data, pj_find.Listing):
             return None
 
         for acc in self.data.accommodation_infos or []:
-            if match := self.HOTEL_STARS_REGEX.match(acc.category):
-                return float(match.group("rating"))
+            if acc.category:
+                if match := self.HOTEL_STARS_REGEX.match(acc.category):
+                    return float(match.group("rating"))
```
remi-dupre, line 580:
```suggestion
            if acc.category and (match := self.HOTEL_STARS_REGEX.match(acc.category)):
                return float(match.group("rating"))
```

We could also have done that, but this `:=` operator quickly makes things weird and harder to read, I'm not sure what to think of it ^^'

remi-dupre (approved):

PR #274, opened 2021-10-06
[QMAPS-2039] Normalized PJ addresses

fatal69100:
- Transform uncomplete word such as `r -> rue` or `bd -> boulevard` 
- Normalize capitalization for each PJ endpoints


fatal69100:
I totally agree by using `Squash and Merge` feature. About avoiding Github tools, do you mean using only cli or can I click on the big green button ?

PR #280, opened 2021-11-03
Update http response for `search` endpoint

fatal69100:
Fix https://jira.qwant.ninja/browse/QMAPS-2199

Return 204 http response when the query is acceptable but no results are return from elasticsearch
Return 200 http response when the query is acceptable and results are return from elasticsearch OR the query is empty and no result from ES. 
We also update the query field with the actual query made to ES

### DIFF
--- a/idunn/api/search.py
+++ b/idunn/api/search.py
@@ -29,8 +29,8 @@ async def search(
         return Response(status_code=200, content=build_empty_query_content_response())
     response = await get_autocomplete(query, extra)
 
-    # When no result was found for an acceptable query
-    if not response.features and query.q != "":
+    # When no result was found for an acceptable query (not empty)
+    if not response.features:
         return Response(status_code=204)
 
     # When an intention is detected, it takes over on geocoding features
--- a/idunn/api/search.py
+++ b/idunn/api/search.py
@@ -6,7 +6,7 @@ from idunn import settings
 from idunn.api.geocoder import get_autocomplete
 from idunn.utils.result_filter import ResultFilter
 from idunn.instant_answer import normalize
-from ..geocoder.models import ExtraParams, QueryParams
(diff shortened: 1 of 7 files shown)


idunn/api/search.py
```diff
@@ -26,6 +27,10 @@ async def search(
     query.q = normalize(query.q)
     response = await get_autocomplete(query, extra)
 
+    # When no result was found for an acceptable query
+    if len(response.dict()["features"]) == 0 and query.q != "":
```
remi-dupre, line 31: This should work:

```suggestion
    if response.features and query.q != "":
```

idunn/api/search.py
```diff
@@ -26,6 +27,10 @@ async def search(
     query.q = normalize(query.q)
     response = await get_autocomplete(query, extra)
```
remi-dupre, line 28: This shouldn't append often but it is a quite inefficient that we still call the geocoder when `q == ""`. You still check if `response.features` is empty, but I don't think there is any case where we expect the geocoder to give a result?

idunn/api/search.py
```diff
@@ -24,8 +25,14 @@ async def search(
     """
     # Fetch autocomplete results which will be filtered
     query.q = normalize(query.q)
+    if query.q == "":
+        return Response(status_code=200, content=build_empty_query_content_response())
     response = await get_autocomplete(query, extra)
 
+    # When no result was found for an acceptable query
+    if not response.features and query.q != "":
```
remi-dupre, line 33: Here `query.q` will never be empty, otherwise we would have early-returned a few lines above.

I have mixed feelings about this, so if this is intentional for the sake of explicitness I'm ok with it. Otherwise a comment would probably do the work :)

remi-dupre (dismissed):

remi-dupre (approved):

PR #282, opened 2021-11-09
Remove legacy pj_filters

fatal69100:


remi-dupre (approved):
This should be the last of it. Bye bye again random pagesjaunes index :wave:

PR #286, opened 2021-11-22
Gunicorn

sinaure:
1. fix readonly issue caused by pid file
2. allow image build and push for every branch

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,11 +1,4 @@
-on:
-  pull_request:
-  push:
-    branches:
-      - 'master'
-      - 'es7'
-    tags:
-      - 'v*'
+on: [push, pull_request]
 
 env:
   DOCKER_IMAGE_BASENAME: qwantresearch/idunn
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yaml
```diff
@@ -16,28 +9,40 @@ jobs:
...
-
       - run: docker build --label "org.label-schema.vcs-ref=$GITHUB_SHA" -t $DOCKER_IMAGE .
-
-      - if:  github.event_name == 'push'
-        run: docker push $DOCKER_IMAGE
+      
+      - name: Docker push
+        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        run: |
+          docker push $DOCKER_IMAGE
```
remi-dupre, line 47: Is that a good practice to push a docker image for all branches by default? I feel like we will be polluting Docker Hub a lot.

In the case of Idunn this is not a big issue as the last layers mostly contain Python sources, so we only upload a few megabytes. But on other repositories the cost can be couples of dozens of megabytes if there are compiled binaries or big raw data in the image. 

The only reason why this could look ok to me in a more general situation is that we are using a free Docker Hub account for which [inactive images appears to be removed after 6 months](https://github.com/docker/roadmap/issues/152).

remi-dupre (approved):

.github/workflows/docker.yaml
sinaure: is a good practice to test a feature before merge to master

PR #288, opened 2021-11-30
Refacto datasources with a factory

fatal69100:
- [x] Trigger Tripadvisor sources in specific cases for Maps : https://confluence.qwant.ninja/confluence/pages/viewpage.action?pageId=48772082.
- [x] Refacto Datasources in a more OOP paradigm to easily maintain multiples datasources
- [x] Refacto two `utils.py` module files used in multples modules into utils modules


idunn/api/places_list.py
```diff
@@ -228,33 +190,61 @@ async def get_places_bbox_impl(
...
+
+async def select_datasource_outside_france(params):
+    if any(cat in params.category for cat in TRIPADVISOR_CATEGORIES_COVERED_WORLDWIDE):
+        params.source = PoiSource.TRIPADVISOR
+    else:
+        params.source = PoiSource.OSM
+
+
+async def is_brand_detected_or_pj_category_cover(params):
+    return params.q or (params.category and all(c.pj_what() for c in params.category))
```
remi-dupre, line 217: This doesn't require to be async, which implies that `select_datasource_*` either

idunn/api/places_list.py
```diff
@@ -228,33 +190,61 @@ async def get_places_bbox_impl(
...
+        await select_datasource_for_france(params)
+    else:
+        await select_datasource_outside_france(params)
 
-    # Default source (OSM) with category or class/subclass filters
-    if params.raw_filter:
-        filters = [MimirPoiFilter.from_url_raw_filter(f) for f in params.raw_filter]
+
+async def select_datasource_for_france(params):
+    if any(cat in params.category for cat in TRIPADVISOR_CATEGORIES_COVERED_IN_FRANCE):
```
remi-dupre, line 201: I'd argue that this `any` should be `all` if we want to be sure to cover all categories with the selected datasource. I don't think it actually matters because for a regular use of Qwant Maps this `category` parameter should always be a list of exactly one element?

idunn/api/places_list.py
```diff
@@ -228,33 +190,61 @@ async def get_places_bbox_impl(
...
+    def get_datasource(self, source_type: PoiSource) -> Datasource:
+        """Get the matching datasource to fetch POIs"""
+        if source_type == PoiSource.TRIPADVISOR:
+            return Tripadvisor()
+        if source_type == PoiSource.PAGESJAUNES:
+            return PagesJaunes()
+        if source_type == PoiSource.OSM:
+            return Osm()
+        logger.error("%s is not a valid source type, OSM source is used as fallback", source_type)
+        return Osm()
```
remi-dupre, line 250: I think we should crash the function as this could only happen if there is a critical issue in our implementation.

I would reserve this kind of behavior (error log + fallback) for cases where an external API returns an unexpected result.

idunn/api/places_list.py
fatal69100: If we use `all` to cover all categories, I think that I should reverse the expression : `all(cat in TRIPADVISOR_CATEGORIES_COVERED_IN_FRANCE for cat in params.category)`

idunn/api/places_list.py
fatal69100: Definitely

PR #290, opened 2021-12-02
Trigger instant answer with tripadvisor 

fatal69100:
Basically, when Instant Answers find no intention : call asynchronously Tripadvisor, PJ and OSM. If we have a hotel with Tripadvisor, we return it. If not, we return PJ answers. If there are no answers, return OSM.

- [x] Add tests to fit https://confluence.qwant.ninja/confluence/pages/viewpage.action?pageId=48772082
- [x] Use of asyncio to asynchronously trigger searchs

### DIFF
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -266,10 +266,10 @@ async def get_instant_answer(
             intentions[0].description._place_in_query,
         )
 
-    # Query PJ API asynchronously as a task which may be cancelled if bragi
-    # finds results.
+    # Query PJ API and Bragi osm asynchronously as a task which may be cancelled
+    # if bragi tripadvisor finds results.
     fetch_pj = asyncio.create_task(fetch_pj_response())
-    bragi_osm_response = await bragi_client.autocomplete(query)
+    fetch_bragi_osm = asyncio.create_task(bragi_client.autocomplete(query))
     bragi_tripadvisor_response = await bragi_client.autocomplete(query_tripadvisor)
     bragi_tripadvisor_features = result_filter.filter_bragi_features(
         normalized_query, bragi_tripadvisor_response["features"]
@@ -282,6 +282,7 @@ async def get_instant_answer(
             and feature_properties["poi_type"]["id"].split(":")[1] == "subclass_hotel"
         ):
             fetch_pj.cancel()
+            fetch_bragi_osm.cancel()
             place_id = feature_properties["id"]
             return await run_in_threadpool(
                 get_instant_answer_single_place,
@@ -291,19 +292,10 @@ async def get_instant_answer(
                 type="poi-tripadvisor",
             )
 
-    bragi_osm_features = result_filter.filter_bragi_features(
-        normalized_query, bragi_osm_response["features"]
-    )
-    if bragi_osm_features:
-        fetch_pj.cancel()
-        place_id = bragi_osm_features[0]["properties"]["geocoding"]["id"]
-        return await run_in_threadpool(
-            get_instant_answer_single_place, query=q, place_id=place_id, lang=lang
-        )
-
     pj_response = result_filter.filter_places(normalized_query, await fetch_pj)
 
     if pj_response:
+        fetch_bragi_osm.cancel()
         place_id = pj_response[0].get_id()
         result = InstantAnswerResult(
             places=[pj_response[0].load_place(lang)],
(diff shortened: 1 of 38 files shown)


idunn/places/poi.py
```diff
@@ -80,6 +77,19 @@ def get_contribute_url(self):
         except ValueError:
             return None
 
+    def get_source(self):
+        return PoiSource.OSM
+
+
+class OsmPOI(POI):
+    def get_source(self):
+        return PoiSource.OSM
```
remi-dupre, line 86: This seems like an equivalent to the base class `Poi`. Maybe that's this base class that should be renamed into `OsmPOI`? Otherwise we could make `Osm` into an abstract class to explicitly deprecate direct use and avoid confusion.

idunn/utils/result_filter.py
```diff
@@ -340,7 +340,8 @@ def build_params_from_bragi(cls, query: str, bragi_response: dict) -> dict:
         local_names = [
             prop["value"]
             for prop in bragi_response.get("properties", [])
-            if prop["key"].startswith("name:") or prop["key"].startswith("alt_name:")
+            if isinstance(prop, dict)
```
remi-dupre, line 343: When is this happening? The field may be `null`?

fatal69100: This is happening on `properties` of Tripadvisor POI 
```
  "properties": {
    "ta:average_rating": "4.5",
    "ta:photos_url": "https://www.tripadvisor.com/Restaurant_Review-g1743691-d3166925-Reviews-Chez_Eric-Vaucluse_Provence_Alpes_Cote_d_Azur.html?m=66562#photos",
    "ta:review_count": "321",
    "ta:url": "https://www.tripadvisor.com/Restaurant_Review-g1743691-d3166925-Reviews-Chez_Eric-Vaucluse_Provence_Alpes_Cote_d_Azur.html?m=66562",
    "website": "http://www.facebook.com/profile.php?id=778236518876620",
    "image": "https://media-cdn.tripadvisor.com/media/photo-o/0f/e9/04/82/photo0jpg.jpg"
  }
```
is different from OSM POI : 
```
  "properties": [
    {
      "key": "addr:street",
      "value": "Rue de la Légion d'Honneur"
    }
 ]
```

idunn/places/poi.py
fatal69100: Ok we can use OSM as an abstract class.  there are three type : OSMPoi, TripadvisorPoi and BragiPoi. The weird thing is that BragiPoi can be a OSMPoi or a TripadvisorPoi. I am not really happy with what i did because everything is mix up

idunn/api/instant_answer.py
```diff
@@ -288,4 +305,15 @@ async def fetch_pj_response():
...
+    bragi_osm_response = await fetch_bragi_osm
+    bragi_osm_features = result_filter.filter_bragi_features(
+        normalized_query, bragi_osm_response["features"]
+    )
+
+    if bragi_osm_features:
+        place_id = bragi_osm_features[0]["properties"]["geocoding"]["id"]
+        return await run_in_threadpool(
+            get_instant_answer_single_place, query=q, place_id=place_id, lang=lang
+        )
```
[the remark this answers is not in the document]
fatal69100: reversed

idunn/api/places_list.py
```diff
@@ -223,7 +223,7 @@ async def _fetch_extended_bbox(bbox_extended, params, places_list):
     original_bbox_height = original_bbox[3] - original_bbox[1]
     original_bbox_size = max(original_bbox_height, original_bbox_width)
     if original_bbox_size < EXTENDED_BBOX_MAX_SIZE:
-        # Compute extended bbox and fetch results a second time
+        # Compute extended bbox and fetch results a second time_fetch_extended_bbox
```
remi-dupre, line 226: That's probably a refactor gone wrong

idunn/places/poi.py
```diff
@@ -13,6 +13,10 @@ class POI(BasePlace):
 
     def __init__(self, d):
         super().__init__(d)
+        if self["id"].startswith("ta"):
```
remi-dupre, line 16: I would even be extra careful in case we have another ID namespace which start with "ta" someday:

```suggestion
        if self["id"].startswith("ta:"):
```

remi-dupre (approved):

PR #292, opened 2021-12-16
Add MCID tripadvisor url

fatal69100:
Add MCID tripadvisor url in `PlaceMeta` object with `source_url` and `contribute_url` fields

Notice, I keep PlaceMeta object for Tripadvisor even if `contribute_url` doesn't make sense for Tripadvisor. I can change if neccessary

### DIFF
--- a/idunn/places/poi.py
+++ b/idunn/places/poi.py
@@ -4,6 +4,7 @@ from urllib.parse import urlencode
 from idunn import settings
 from idunn.api.constants import PoiSource
 from .base import BasePlace
+from abc import abstractmethod
 
 OSM_CONTRIBUTION_HASHTAGS = settings["OSM_CONTRIBUTION_HASHTAGS"]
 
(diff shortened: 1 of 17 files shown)


idunn/places/poi.py
```diff
@@ -82,7 +78,49 @@ def get_contribute_url(self):
             return None
 
     def get_source(self):
-        return self.source
+        pass
+
+
+class OsmPOI(POI):
+    @cached_property
```
remi-dupre, line 85: I'd argue that a `.rsplit` probably doesn't cost enough to justify using `@cached_property` (which could lead to surprising behaviors), but it wasn't introduced here anyway :3

idunn/places/poi.py
```diff
@@ -82,7 +78,49 @@ def get_contribute_url(self):
...
+
+    def get_contribute_url(self):
+        try:
+            osm_kind, osm_id = self.osm_id_tuple
+            edit_params = {osm_kind: osm_id}
+            if OSM_CONTRIBUTION_HASHTAGS:
+                edit_params["hashtags"] = OSM_CONTRIBUTION_HASHTAGS
+            return f"https://www.openstreetmap.org/edit?{urlencode(edit_params)}"
+        except ValueError:
+            return None
```
remi-dupre, line 109: I think your intention was to also remove their implementation from the class `OsmPoi`?

remi-dupre (changes requested):

remi-dupre:
> Notice, I keep PlaceMeta object for Tripadvisor even if `contribute_url` doesn't make sense for Tripadvisor. I can change if neccessary

I don't think it is an issue as it is already optional [for PJ POIs](https://github.com/Qwant/idunn/blob/es7/idunn/places/pj_poi.py#L317).

remi-dupre (approved):

PR #293, opened 2022-01-05
Bragi new params

remi-dupre:
Update Bragi API to latest changes, mostly poi dataset name (https://github.com/CanalTP/mimirsbrunn/pull/668).

### DIFF
--- a/idunn/geocoder/models/params.py
+++ b/idunn/geocoder/models/params.py
@@ -30,54 +30,70 @@ class Type(str, Enum):
     Zone = "zone"
 
 
+class PlaceDocType(str, Enum):
+    # pylint: disable=invalid-name
+    Admin = "admin"
+    Street = "street"
+    Addr = "addr"
+    Poi = "poi"
+    Stop = "stop"
+
+
 @dataclass
 class QueryParams:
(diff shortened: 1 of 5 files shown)


idunn/geocoder/models/params.py
```diff
@@ -30,54 +30,70 @@ class Type(str, Enum):
     Zone = "zone"
 
 
+class PlaceDocType(str, Enum):
+    # pylint: disable=invalid-name
+    Admin = "admin"
+    Street = "street"
+    Addr = "addr"
+    Poi = "poi"
+    Stop = "stop"
```
fatal69100, line 39: I think Idunn should not be aware about `stop` type

fatal69100 (approved):

idunn/geocoder/models/params.py
remi-dupre: This makes sense :+1:

remi-dupre: Do you think we should also remove `StopArea = "public_transport:stop_area"` from the enum `Type`?

fatal69100: yes

PR #300, opened 2022-01-21
add heathcheck for all external sources

fatal69100:
heathcheck for : 
- es_mimir
- es_wiki
- bragi
- taggeur
- classifier
- pagesjaunes
- redis

### DIFF
--- a/idunn/api/closest.py
+++ b/idunn/api/closest.py
@@ -4,7 +4,7 @@ from fastapi import HTTPException
 from pydantic import confloat
 
 from idunn import settings
-from idunn.utils.es_wrapper import get_elasticsearch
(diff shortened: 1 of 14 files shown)


remi-dupre (commented):
A few other APIs come to my mind:
 - Bragi: we were not sure how to check for liveness without risking to query data that is not available for given environment, a very sane solution would be to query its own `status` endpoint (we probably just both forgot it existed).
 - NLP: maybe we should split tagger / classifier? That's actually probably not very important.
 - Wikipedia's API: it's only a non-critical fallback but that's one reason we not notice if there was an issue.

remi-dupre (approved):

PR #302, opened 2022-01-24
update dependencies

remi-dupre:
Similar to https://github.com/Qwant/idunn/pull/301, the ipython dependency has a security breach in currently pined version.

I couldn't produce the exact same result as dependabot, so I just ended up updating everything with a `pipenv update --dev`, I suppose that it can't be a bad idea.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,19 +18,19 @@
     "default": {
         "anyio": {
             "hashes": [
-                "sha256:4fd09a25ab7fa01d34512b7249e366cd10358cdafc95022c7ff8c8f8a5026d66",
-                "sha256:67da67b5b21f96b9d3d65daa6ea99f5d5282cb09f50eb4456f8fb51dffefc3ff"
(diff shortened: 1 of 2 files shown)


idunn/api/urlsolver.py
```diff
@@ -39,7 +39,7 @@ async def follow_redirection(
     if not hmac.compare_digest(hash_url(url), hash):
         raise HTTPException(403, detail="provided hash does not match the URL")
 
-    response = await client.get(url, allow_redirects=False)
+    response = await client.get(url, follow_redirects=False)
```
remi-dupre, line 42: This parameter doesn't seem to have changed. I think this was a mistake, but the default value is actually `False` so it had no impact (still better to be explicit here). Either *pylint* go better or *httpx* has a better type annotation.

Cf. https://www.python-httpx.org/compatibility/#redirects

fatal69100 (approved):
outch.. there are conflicts

PR #305, opened 2022-01-31
Qmaps 2453 fix TA module priority

fatal69100:


idunn/api/instant_answer.py
```diff
@@ -267,53 +274,79 @@ async def fetch_pj_response():
...
-            return await run_in_threadpool(
-                get_instant_answer_single_place,
-                query=q,
-                place_id=place_id,
-                lang=lang,
-                type="poi_tripadvisor",
-            )
-
+    # Select datasource instant answer in France
+    if len(intentions) > 0 and pj_source.bbox_is_covered(intentions[0].filter.bbox):
```
remi-dupre, line 285: I'm not sure I understand the reason for `len(intentions) > 0`. Maybe it is put here in order to check if the user is actually looking for an hotel?

idunn/api/instant_answer.py
```diff
@@ -267,53 +274,79 @@ async def fetch_pj_response():
...
+                return build_response(result, query=q, lang=lang)
+    # Call tripadvisor datasource instant answer outside France or
+    # without intention location detection
+    else:
+        for bragi_tripadvisor_feature in bragi_tripadvisor_features:
+            feature_properties = bragi_tripadvisor_feature["properties"]["geocoding"]
+            if (
+                "poi_types" in feature_properties
+                and feature_properties["poi_types"][0]["id"].split(":")[0]
+                in AVAILABLE_CLASS_TYPE_TRIPADVISOR
```
remi-dupre, line 323: I feel like there is some code duplication with the french part, maybe we could build a list `allowed_classes_types` which would be either `AVAILABLE_CLASS_TYPE_TRIPADVISOR` or `["class_hotel"]` depending on the location?

idunn/api/instant_answer.py
fatal69100: I check before accessing `intentions[0]` to prevent 'NoneType' AttributeError

idunn/api/instant_answer.py
fatal69100: Indeed there is some code duplication, but I can't easily refactor with interruption workflow in the loop. Probably, need another PR to refactor more largely instant answer

PR #307, opened 2022-02-17
Add TRIPADVISOR_ENABLED flag

fatal69100:


remi-dupre (changes requested):
As we've cleaned-up `categories.yml` for the categories that won't be covered by PagesJaunes anymore (that's only hotels ?), setting `IDUNN_TRIPADVISOR_ENABLED=False` will search hotels from OSM and not from PagesJaunes. I suppose we just have to restore the `pj_what` field for hotels.

remi-dupre (approved):
LGTM :+1:

PR #308, opened 2022-02-18
wrap tripadvisor's ratings image in thumbr

remi-dupre:
Rating image URLs currently expose the client to an external website, this PR wraps the images with the thumbr to keep the client's IP & cie safe.

See https://github.com/Qwant/erdapfel/pull/1278

### DIFF
--- a/idunn/places/base.py
+++ b/idunn/places/base.py
@@ -283,9 +283,11 @@ class BasePlace(dict):
 
     def get_bubble_star_url(self):
         if self.properties.get("ta:average_rating") is not None:
+            rating = float(self.properties.get("ta:average_rating"))
+
             base_url = (
                 f"https://www.tripadvisor.com/img/cdsi/img2/ratings/traveler/"
-                f"{self.properties.get('ta:average_rating')}-MCID-66562.svg"
+                f"{rating:.1f}-MCID-66562.svg"
             )
 
             if thumbr.is_enabled():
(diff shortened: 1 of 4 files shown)


idunn/places/base.py
```diff
@@ -282,10 +283,18 @@ def get_meta(self):
 
     def get_bubble_star_url(self):
         if self.properties.get("ta:average_rating") is not None:
-            return (
+            rating = float(self.properties.get("ta:average_rating"))
+
+            base_url = (
                 f"https://www.tripadvisor.com/img/cdsi/img2/ratings/traveler/"
-                f"{self.properties.get('ta:average_rating')}-MCID-66562.svg"
+                f"{rating:.1f}-MCID-66562.svg"
```
remi-dupre, line 290: There was a small bug here as https://www.tripadvisor.com/img/cdsi/img2/ratings/traveler/4.5-MCID-66562.svg is a valid URL while https://www.tripadvisor.com/img/cdsi/img2/ratings/traveler/4-MCID-66562.svg isn't. So we have to ensure that the rating is formatted with exactly one digit after the decimal point.

fatal69100: We can add a comment above for this trick : `Tripadvisor bubble star url need a rating with a decimal (e.g 4.0 or 4.5)`

fatal69100: Nice catch btw :+1:

fatal69100 (approved):

PR #310, opened 2022-03-09
patterns_to_ignore: add contacts?

remi-dupre:
This should help with requests like "contact de %POI" "%POI contacts",
etc...


fatal69100 (changes requested):
Maps should deserve several `contact` request. The fix should be done in the NLP block.

remi-dupre:
After discussion, this could create regressions for few benefits

PR #319, opened 2022-03-18
Clean Instant answer old flag

fatal69100:
Remove : `IA_SUCCESS_ON_GENERIC_QUERIES` and `IA_CALL_PJ_POI` flag


remi-dupre (approved):
I'm sorry, I've introduced some conflicts while you were off :grimacing: 

Do you think we should also remove some very technical settings, for example I have very recently introduced BRAGI_MAX_CONNECTIONS in https://github.com/Qwant/idunn/pull/320, but it was a temporary way to prevent the application from crashing while investigating https://github.com/Qwant/idunn/pull/323

fatal69100:
we can keep it, it could be still useful

remi-dupre (approved):

PR #324, opened 2022-03-24
Add public transports ft. Hove's API

remi-dupre:
Add public transport support through Hove's API.

This is built by building a translation layer that will convert Hove's API result into a format mostly compatible with Mapbox (a few structures are extended with new fields), which can be interpreted by Erdapfel seamlessly and in the same fashion as with former Combigo API.

I also took my change to move every directions-related stuff in `idunn.datasources.directions`, which means a bit of refactoring in the mapbox client.


idunn/datasources/hove/models.py
```diff
@@ -0,0 +1,338 @@
+"""
+Derialize Hove's responses and conversion primitives following the following equivalence:
```
fatal69100, line 2: following the following :p (maybe following the next equivalences)

idunn/datasources/hove/models.py
```diff
@@ -0,0 +1,338 @@
...
+class Instruction(BaseModel):
+    direction: int
+    duration: int
+    instruction: Optional[str]
+    instruction_start_coordinate: Optional[Coordinate]
+    length: int
+    name: str
+
+    def get_api_modifier(self) -> str:
+        # See https://docs.api.com/api/navigation/directions/#step-maneuver-object
```
fatal69100, line 53: Should `https://docs.api.com` be replaced by the actual hove documentation URL ?

idunn/datasources/hove/models.py
```diff
@@ -0,0 +1,338 @@
...
+            if not inst.instruction_start_coordinate:
+                continue
+
+            end_coord = inst.instruction_start_coordinate.to_position()
+
+            try:
+                end_index = next(
+                    i
+                    for i, coord in enumerate(all_coords)
+                    if abs(coord[0] - end_coord[0]) < 1e-6 and abs(coord[1] - end_coord[1]) < 1e-6
```
fatal69100, line 155: 1e-6 is something like 1meter ?

idunn/datasources/hove/models.py
```diff
@@ -0,0 +1,338 @@
...
+            legs=legs,
+            geometry=geometry,
+            summary=summary,
+        )
+
+
+class HoveResponse(BaseModel):
+    journeys: List[Journey] = []
+
+    def as_api_response(self) -> api.DirectionsResponse:
```
fatal69100, line 334: why is there a `as` in this method name and in the other ones? Is it to specify that this method is a conversion to the qwant direction model response ?

idunn/datasources/hove/models.py
```diff
@@ -0,0 +1,338 @@
...
+                            else {}
+                        ),
+                    },
+                    "geometry": {
+                        "type": "MultiLineString",
+                        "coordinates": [step.geometry["coordinates"] for step in leg.steps],
+                    },
+                }
+                for sec, leg in zip(secs, legs)
+                for step in leg.steps
```
fatal69100, line 316: Maybe it could be deleted or add the behavior in another pr ?

idunn/datasources/hove/client.py
```diff
@@ -10,6 +10,7 @@
 DIRECT_PATH_MAX_DURATION = 86400  # 24h
 MIN_NB_JOURNEYS = 2
 MAX_NB_JOURNEYS = 5
+FREE_RADIUS = 50  # meters
```
fatal69100, line 13: I try to understand https://doc.navitia.io/?shell#journeys : we don't calculate the time between the given coordinate and the journey departure/arrival in a radius of 50m ?

fatal69100 (approved):
LGTM, some minor question because I was far from this topic

idunn/datasources/hove/client.py
remi-dupre: Yeah exactly, in my understanding it will help with not having weird shenanigans where the itinerary will go around the starting building or take a weird exit from the arrival.

It will affect the computed duration so it shouldn't be more than 50m, 50m meters might already be too much as it means we consider that the user can teleport for a distance of 100m overall.

idunn/datasources/hove/models.py
remi-dupre: Oh maybe this is misleading, leg is not specific to walking sections, this is actually roughly the equivalent of a section in the mapbox API. So this part of the code is responsible for constructing the lines that will be displayed on the map.

Screenshot_2022-07-05_10-00-15

idunn/directions/client.py
```diff
@@ -28,29 +29,29 @@ class MapboxAPIExtraParams(BaseModel):
...
+            case IdunnTransportMode.BIKE:
+                return methods[settings["DIRECTIONS_PROVIDER_CYCLE"]]
+            case IdunnTransportMode.WALKING:
+                return methods[settings["DIRECTIONS_PROVIDER_WALK"]]
+            case IdunnTransportMode.PUBLICTRANSPORT:
+                return methods[settings["DIRECTIONS_PROVIDER_PUBLICTRANSPORT"]]
 
     @staticmethod
     def is_in_allowed_zone(mode: str, from_place: BasePlace, to_place: BasePlace):
+        print(settings["PUBLIC_TRANSPORTS_RESTRICT_TO_CITIES"])
```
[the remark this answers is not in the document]
remi-dupre: :see_no_evil:

idunn/datasources/hove/client.py
```diff
@@ -0,0 +1,68 @@
...
+        self.api_url = settings["HOVE_API_BASE_URL"]
+        self.session = httpx.AsyncClient(verify=settings["VERIFY_HTTPS"])
+        self.session.headers["User-Agent"] = settings["USER_AGENT"]
+
+    async def directions(
+        self,
+        start: BasePlace,
+        end: BasePlace,
+        mode: IdunnTransportMode,
+        _lang: str,
```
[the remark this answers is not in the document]
remi-dupre: From what I understand Navitia only returns localized string targeting the country were the itinerary is, so this parameter will probably be ignored :/ We need the function to take the same parameters as the mapbox one but maybe it should be part of extra?

idunn/datasources/hove/models.py
remi-dupre: :vomiting_face:

idunn/api/geocoder.py
```diff
@@ -7,6 +7,7 @@
 from ..geocoder.bragi_client import bragi_client
 from ..geocoder.models.geocodejson import Intention
 from ..geocoder.nlu_client import nlu_client, NluClientException
```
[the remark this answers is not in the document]
remi-dupre: Indeed!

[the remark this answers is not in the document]
fatal69100: Ok it think that it's quite standard

idunn/datasources/hove/client.py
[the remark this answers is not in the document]
fatal69100: Not necessary. the only thing is, if i search in Italia I will get Italian instruction ? Maybe it's enough for a first version

idunn/datasources/hove/models.py
fatal69100: I am not sure to understand. The `step` variable is not used in the code

idunn/datasources/hove/models.py
remi-dupre: Oh your right, that should be a documented constant, I put it here during the POC and forgot it after that.

Worst case scenario, for the latitude a variation of $1\mathrm{e}{-6}°$ leads to a variation of $$\frac{1\mathrm{e}{-6}}{180} \cdot 40 \cdot 10^{6} \approx 0.22m$$

This is mainly intended to deal with rounding errors, but 22cm seems like a fair margin :shrug:

idunn/datasources/hove/models.py
remi-dupre: Yeah, this is the idea I wanted to convey with the `as_api_` prefix.

As you are asking this I suppose it doesn't do a great job of conveying this idea ? :cry:

idunn/datasources/hove/models.py
remi-dupre: Oh yes indeed, that's a mistake as we already have this:

```python
"coordinates": [step.geometry["coordinates"] for step in leg.steps],
```

This is due to an older version of the code that I figured out could be simplified with a *MultiLineString*, but I forgot to remove line :sob:

idunn/datasources/hove/models.py
remi-dupre: :laughing:

remi-dupre: It to me a minute to understand what happened here, the actual link is https://docs.mapbox.com/api/navigation/directions/#step-maneuver-object

> I'll search and replace "mapbox" -> "api", what could go wrong here?

I'll fix this and add some doc to explain the role of this function.

remi-dupre:
@fatal69100 I've updated the PR, following your review advises. Mostly:

 - moved out Python10 / dep upgrades in https://github.com/Qwant/idunn/pull/365 and https://github.com/Qwant/idunn/pull/366
 - fixed existing test (that was mostly due to the fact that I changed the calls to mapbox to async calls, so the mocks had to be adapted)
 - refactored all mapbox-related stuff in its whole module, in *idunn.datasources.directions.mapbox*
 - added a bit of testing in *test_directions_hove.py*
 - added conversion for some missing fields (summary, currency, price, ...)
 
I think I can consider this branch to be in a "clean" state, so I'll probably work in new PRs for former changes. The fact that it takes 10s for git to rebase this 53 commits history makes it obvious that this branch has been kept alive for way too long.

So feel free to get a second pass of review when you have time, I'm sorry for not keeping this PR more concise :grimacing:

fatal69100:
LGTM :) gg

PR #325, opened 2022-03-24
Add lru_cache on check_ban endpoint call

fatal69100:
When we are scrapped, we reduce the number of call to the check_api endpoint.
We also reduce the number of call for an active user

### DIFF
--- a/idunn/utils/ban_check.py
+++ b/idunn/utils/ban_check.py
@@ -1,4 +1,5 @@
 import logging
+from functools import lru_cache
 from typing import Optional
 from httpx import AsyncClient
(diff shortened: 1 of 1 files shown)


remi-dupre (changes requested):
I think this is not a good idea in the end.

Using std's lru cache won't do the trick as a scrapper would keep himself on top of the cache and we wouldn't renew the eventual negative answer from the antiscrap API that we received after his very first request. This could be fixed by adding a maximum age to the keys in the cache or using something more naive, but I don't think it deserve the pain as I was told that the API we hit on is using a cache.

PR #326, opened 2022-03-24
Remove starlette dependency

fatal69100:
Starlette dep is not required anymore and could be replaced


remi-dupre (dismissed):
Nice catch !

remi-dupre (approved):

PR #337, opened 2022-04-15
Remove review block pydantic validator

fatal69100:
Pydantic try to validate other block with this validation annotation that make Idunn not working in int


remi-dupre (approved):
Ah yeah, also feels more explicit in this fashion :+1:

PR #339, opened 2022-04-20
Call OSM on single Instant answer for admin or poi with wikidata information 

fatal69100:
This PR should get result for 'tour Eiffel', 'tour de pise' from OSM for instant answer

### DIFF
--- a/idunn/api/instant_answer.py
+++ b/idunn/api/instant_answer.py
@@ -340,8 +340,8 @@ async def get_instant_answer(
                     "get_instant_answer: Failed to get place with id '%s'", place_id, exc_info=True
                 )
                 return no_instant_answer()
-            # The default value 'True' is used to validate the condition for admin that don't have poi_sublass key in
-            # their dict.
+            # The default value 'True' is used to validate the condition for admin
+            # that don't have poi_sublass key in their dict.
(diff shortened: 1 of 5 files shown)


idunn/api/instant_answer.py
```diff
@@ -323,9 +327,25 @@ async def fetch_pj_response():
...
+        if bragi_osm_response["features"]:
+            place_id = bragi_osm_response["features"][0]["properties"]["geocoding"]["id"]
+            try:
+                place = place_from_id(place_id, lang, follow_redirect=True)
+            except Exception:
+                logger.warning(
+                    "get_instant_answer: Failed to get place with id '%s'", place_id, exc_info=True
+                )
+                return no_instant_answer()
+            if place.wikidata_id and place.properties.get("poi_subclass", True) not in [
```
remi-dupre, line 343: I'm not sure why `True` is the default value in `place.properties.get("poi_subclass", True)`? Leaving the default `None` would be sufficient?

remi-dupre (approved):

idunn/api/instant_answer.py
fatal69100: The default value is used to validate the condition for `admin` that don't have `poi_sublass` key in their dict. I am adding a comment because it is totally implicit.

PR #340, opened 2022-04-22
rate_limiter: use a single connection pool

remi-dupre:
This should help hitting less heavily on the Redis connection pool.

I've also taken this chance to add a rate limit on the place endpoint, as it can easily expose an external API.

### DIFF
--- a/idunn/utils/default_test_settings.yaml
+++ b/idunn/utils/default_test_settings.yaml
@@ -59,7 +59,7 @@ LIST_PLACES_RL_MAX_REQUESTS: 100 # req per client
 LIST_PLACES_RL_EXPIRE: 900 # seconds
 
 ## Place detail
-GET_PLACE_RL_MAX_REQUESTS: 30 # req per client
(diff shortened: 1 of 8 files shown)


idunn/utils/default_test_settings.yaml
```diff
@@ -58,6 +58,10 @@ LIST_PLACES_EXTENDED_BBOX_MAX_SIZE: "0.4" # Lat/lon degrees
 LIST_PLACES_RL_MAX_REQUESTS: 100 # req per client
 LIST_PLACES_RL_EXPIRE: 900 # seconds
 
+## Place detail
+GET_PLACE_RL_MAX_REQUESTS: 30 # req per client
```
fatal69100, line 62: I think 30 is a low limit

fatal69100 (dismissed):

idunn/utils/default_test_settings.yaml
remi-dupre: Let's make it a 60 then!

fatal69100 (approved):

PR #342, opened 2022-04-25
cache calls to isban API

remi-dupre:
Add a cache on the `isban` API in order to consume as little resource as possible for scrappers. This cache must expire after some time in order to not keep a scrapper on top of the LRU cache while the state of "isban" has changed.

Python ecosystem was quite deceiving, so I opted for a custom implementation in utilities (~50 lines of code without comments seems reasonable):
 - standard library's `lru_cache` doesn't handle async and requires an extra wrapper to handle timed expiration
 - I couldn't find any relevant library for this, [asyncio-cache](https://pypi.org/project/async-cache/) could almost do trick but seems a bit overkill for our usecase and doesn't seem to handle any sized mechanism.


tests/test_cache.py
```diff
@@ -126,3 +128,61 @@ def fake_get(*args):
         resp = response.json()
         assert len(mock_wikipedia_response.calls) == 0
         assert not has_wiki_desc(resp)
+
+
+@pytest.mark.asyncio
+async def test_lru_cache_with_expiration():
```
fatal69100, line 134: :smile:

fatal69100 (approved):

PR #345, opened 2022-04-28
Refacto instant answer module

fatal69100:
The goal of the refacto is to dispatch common logic in the different datasources and to filter data in the search request instead of in the results


fatal69100:
Child pr : https://github.com/Qwant/idunn/pull/346

idunn/instant_answer/__init__.py
```diff
@@ -1 +0,0 @@
-from .normalization import normalize
```
remi-dupre, line 1: :sweat_smile:

idunn/services/instant_answer/normalization.py
```diff
@@ -8,7 +8,16 @@
...
+    query = query.strip().lower()
+    query = ignore_pattern_start.sub("", query)
+    query = ignore_pattern_end.sub("", query)
+    if user_country is not None:
+        user_country = user_country.lower()
+    return query, user_country
+
+
+def normalize_search(query) -> str:
     query = query.strip().lower()
```
remi-dupre, line 21: Feels like `idunn.services.**instant_answer**` is not the place then? :/

How would you define what a service is? Maybe we just shouldn't have an `instant_answer` submodule?

idunn/api/instant_answer.py
```diff
@@ -191,26 +164,35 @@ async def get_instant_answer_intention(intention, query: str, lang: str):
...
+        return [
+            (Tripadvisor(), fetch_bragi_tripadvisor),
+            (PagesJaunes(), fetch_pj),
+            (Osm(is_wiki_filter=False), fetch_bragi_osm),
+        ]
+    return [
+        (Osm(is_wiki_filter=True), fetch_bragi_osm),
+        (Tripadvisor(), fetch_bragi_tripadvisor),
+        (Osm(is_wiki_filter=False), fetch_bragi_osm),
+    ]
```
remi-dupre, line 193: Awesome :+1: 

That's a lazy question, but did you have to compromise in order to simplify this?

idunn/datasources/pages_jaunes.py
```diff
@@ -57,6 +60,33 @@ def format_where(bbox):
...
+        if len(intentions) > 0:
+            return asyncio.create_task(
+                run_in_threadpool(
+                    pj_source.search_places,
+                    query,
+                    intentions[0].description._place_in_query,
+                ),
+                name="ia_fetch_pj",
+            )
+        return asyncio.create_task(
```
remi-dupre, line 74: I'd argue that the caller should be responsible of calling `asyncio.create_task` which implies non-trivial side effects. Then we could also make this method async to make it more obvious that it returns an awaitable.

This remark also applies to other `fetch_search` methods.

idunn/datasources/pages_jaunes.py
```diff
@@ -57,6 +60,33 @@ def format_where(bbox):
...
+            run_in_threadpool(
```
remi-dupre, line 75: In a similar fashion, I would have said that it should be left to the caller but maybe we should keep this and make the method async in the `Datasource` parent class.

remi-dupre (changes requested):

idunn/services/instant_answer/normalization.py
fatal69100: I totally give up the service layer in this PR. I think I forgot to undo this move. (I think it will be easier to do it on TS migration)

idunn/api/instant_answer.py
fatal69100: Yes, I have to instantiate a new object for each priority which is not great because there is already some instantiation (for example with pj_source).

idunn/datasources/pages_jaunes.py
fatal69100: yes indeed, i need to change that part

idunn/api/instant_answer.py
remi-dupre: Oh yeah indeed, this means that we will open a new HTTP session with page jaunes for each request.

Maybe we should instantiate those datasources once outside of `get_single_ia_datasource_priority` and only return already existing instances from this function?

remi-dupre (approved):
:heart_eyes:

PR #358, opened 2022-05-24
Call pj on unknown location

fatal69100:
Currently, we don't have a solution to query PagesJaunes for the whole France


fatal69100:
Closed because we can find another strategy

PR #360, opened 2022-05-27
clean combigo direction code

fatal69100:


remi-dupre (approved):
That looks good to me, I think we should merge this as I deleted some parts of it in https://github.com/Qwant/idunn/pull/324 because it would have required some changes anyway.

I'll try to rebase on this and then we should be good :)

PR #365, opened 2022-06-28
Debian os to Alpine

sinaure:
- switch from debian to alpine

- update lockfile


Dockerfile
```diff
@@ -1,27 +1,47 @@
...
+	&&  pip install --no-cache-dir --upgrade pip
+
+WORKDIR /usr/local/src/idunn/src
 
 # Installing packages
-RUN pip install pipenv==2018.11.26
+RUN pip install pipenv
 
-ADD --chown=app_user app.py Pipfile* /app/
+ADD --chown=idunn app.py Pipfile* /usr/local/src/idunn/src/
```
remi-dupre, line 32: Why did we need to change the path here?

On the other side I think this should be this:
```suggestion
ADD --chown=idunn app.py Pipfile* /usr/local/idunn/src/
```

remi-dupre: Also, when trying to run the image *gunicorn* doesn't find the module Idunn, so it shoud probably even be something like this:
```suggestion
ADD --chown=idunn app.py Pipfile* /usr/local/idunn/
```

Dockerfile
```diff
@@ -1,27 +1,47 @@
...
+      idunn
+
+RUN apk update \
+    && apk add --upgrade --no-cache \
+        bash openssh curl ca-certificates openssl less htop gcc git \
+		g++ make wget rsync \
+        build-base libpng-dev freetype-dev libexecinfo-dev openblas-dev libgomp lapack-dev \
+		libgcc libquadmath musl  \
+		libgfortran \
+		lapack-dev \
```
remi-dupre, line 24: How did you build this list of required packages? Some of them like `less` or `htop` seem dubious here :thinking:

Dockerfile
```diff
@@ -1,27 +1,47 @@
...
-RUN apt-get update && apt-get -y install git gcc
-
-RUN useradd -r app_user
-RUN mkdir /app
-RUN chown app_user /app/
-WORKDIR /app
+FROM python:3.10-alpine
+
+ENV PYTHONPATH="/usr/local/src/idunn/src"
+ENV PYTHONUNBUFFERED=1
```
remi-dupre, line 4: As I understand it, it is recommended to disable output buffers in docker images to avoid issues where a log line is not outputted before a pod crashes (I'm not sure how you can do this without raising a Python exception but I'll trust the community on this one) :+1:

Dockerfile
```diff
@@ -1,27 +1,47 @@
-FROM python:3.8-slim
-
-RUN apt-get update && apt-get -y install git gcc
-
-RUN useradd -r app_user
-RUN mkdir /app
-RUN chown app_user /app/
-WORKDIR /app
+FROM python:3.10-alpine
```
remi-dupre, line 1: Is the 3.10 / Pipfile.lock upgrade something related to Alpine / CVE hunting or just some maintenance?

Anyway this is good, I'll take benefits of this to extract this change from the public transports branch :smile:

remi-dupre (commented):
I was willing to finish the work that was started here but I have a few questions first :3

remi-dupre:
@sinaure I've updated your branch to fix runtime issues and to my chances with some cleanup in dependencies (I suppose the goal of using Alpine is to have as little packages as possible in the result image):

 - removed most installed packaged
 - added a builder images which constructs Idunn's dependencies, the final image thus doesn't need to have *pipenv* and other build dependencies installed

I was also wondering why we have to build a specific user? Is default docker user unsafe?

Dockerfile
sinaure: To be consinstend with Filesystem Hierarchy Standard (FHS) /app does not exist

Dockerfile
sinaure: this part I have just copy pasted let's see if we can remove something. The most annyoing dependency was spacy

Dockerfile
sinaure: I am pretty sure that the same pipfile.lock is usable both in debian and Alpine

Dockerfile
```diff
@@ -1,31 +1,59 @@
...
+RUN apk update && apk add --upgrade --no-cache geos
+
+RUN mkdir -p /home/idunn/prometheus_multiproc
+USER idunn
+WORKDIR /home/idunn
+
+# Add files into images
+ADD app.py /home/idunn
+ADD idunn /home/idunn/idunn
+COPY --chown=idunn --from=builder /usr/local/src/.venv /home/idunn/.venv
```
madrisan, line 47: Are you sure this change of ownership is required? Better to leave the binary owned by root (not writeable by the used that will execute the application).

PR #371, opened 2022-07-20
directions/hove: add walk steps and fix distance in summary

remi-dupre:


idunn/datasources/directions/hove/models.py
```diff
@@ -217,7 +217,11 @@ def as_api_route_summary_part(self) -> api.RouteSummaryPart:
                 if self.display_informations
                 else None
             ),
-            distance=42,
```
fatal69100, line 220: Wow, how did I miss that !

fatal69100 (approved):

idunn/datasources/directions/hove/models.py
remi-dupre: I though about checking all TODOs but didn't check for the remaining 42's :see_no_evil:

PR #374, opened 2022-08-03
Add airport category

fatal69100:
Category only use for OSM


remi-dupre (approved):
Probably the most stable tag I've ever seen on taginfo O_o

Screenshot_2022-08-04_09-28-20

